PR #2, opened 2020-03-17
Update cddl dep to 0.5.0

rooooooooob:
Verified generated code against old generated code and it is equal except the changes in comments from fixing the Primtive -> Primitive typo


cddl_test/static/prelude.rs
```diff
@@ -6,9 +6,10 @@ use wasm_bindgen::prelude::*;
 //       instead have it either:
 //  1) generate 1 per each wrapped type
 //  2) don't pass it, instead wrap it/unwrap it automatically
+#[derive(Clone)]
```
rooooooooob: This was meant for the previous commit related to the RustType refactor to support tags.

PR #6, opened 2020-04-07
Address generation

rooooooooob:


rust/src/address.rs
```diff
@@ -0,0 +1,559 @@
...
+    Base(BaseAddress),
+    Ptr(PointerAddress),
+    Enterprise(EnterpriseAddress),
+}
+
+#[wasm_bindgen]
+#[derive(Debug, Clone, Eq, Ord, PartialEq, PartialOrd)]
+pub struct Address(AddrType);
+
+// to/from_bytes() are the raw encoding without a wrapping CBOR Bytes tag
```
SebastienGllmt, line 91: No sure if I understand this comment. `to_bytes` should be exactly the inverse to `bech32`.

rust/src/address.rs
```diff
@@ -0,0 +1,559 @@
...
+            },
+            // enterprise
+            0b0110 | 0b0111 => {
+                AddrType::Enterprise(EnterpriseAddress::new(network, read_addr_cred(4, 1)))
+            },
+            // byron
+            0b1000 => {
+                unimplemented!()
+            },
+            // TODO: return error
```
SebastienGllmt, line 184: TODO

rust/src/fees.rs
```diff
@@ -0,0 +1,216 @@
+use super::*;
+// constants - are these right?
```
SebastienGllmt, line 2: presumably if the test cases pass?

rust/src/fees.rs
```diff
@@ -0,0 +1,216 @@
...
+const ABSTRACT_SIZE_VKEY: usize = 8;//64; // from 
+const ABSTRACT_SIZE_SIG: usize = 13;//64;
+const HL: usize = 4;//16; // really unsure about this one, found https://github.com/input-output-hk/cardano-base/blob/501d91e426ae84ce0ae056be38bd3db594af9fc2/cardano-crypto-class/src/Cardano/Crypto/Hash/MD5.hs#L16
```
SebastienGllmt, line 5: what's the result of this question?

rust/src/fees.rs
```diff
@@ -0,0 +1,216 @@
...
+const HASH_OBJ: usize = 2 + HL;
+// CBOR-encoding specific contants I believe
```
SebastienGllmt, line 7: "I believe" <- do we know this for sure now?

rust/src/fees.rs
```diff
@@ -0,0 +1,216 @@
...
+    println!("cert_size({:?}) = {}", cert, x);
+    x
+}
+
+#[wasm_bindgen]
+pub fn txsize(tx: &TransactionBody, witness: &TransactionWitnessSet) -> usize {
+    // It seems that in Haskell we have Tx = {Body, vkeySigs, msigScripts, metadata}
+    // but in the CDDL we have Body as one then vkeySigs/msigScripts as another
+    // as the data is segwit.
+    // We might want to make our own version that wraps these then.
```
SebastienGllmt, line 75: is this required? What is the downside of not doing this?

rust/src/fees.rs
```diff
@@ -0,0 +1,216 @@
...
+                     + HASH_OBJ * witness.script_witnesses.as_ref().map_or(0, Scripts::len);
+
+    let input_size = LABEL_SIZE + CBOR_TAG + ARRAY_PREFIX
+                   + tx.inputs.len() * (SMALL_ARRAY + UINT + HASH_OBJ);
+    let output_size = LABEL_SIZE + ARRAY_PREFIX
+                    + tx.outputs.len() * (SMALL_ARRAY + UINT + ADDRESS);
+    let certs_size = tx.certs.as_ref().map_or(0, |certs| certs.0.iter().map(cert_size).sum::<usize>());
+    let withdrawals_size = LABEL_SIZE + MAP_PREFIX
+                         + tx.withdrawals.as_ref().map_or(0, |wd| wd.len()) * (UINT + CREDENTIAL);
+    // TODO: updates
```
SebastienGllmt, line 92: TODO? Is there any blocker for this one?

rust/src/fees.rs
```diff
@@ -0,0 +1,216 @@
...
+    let update_size = ARRAY_PREFIX;
+    // TODO: metdata/hash
```
SebastienGllmt, line 94: TODO: probably we should make it more clear than an in-code comment that this library won't transactions containing metadata transactions

rust/src/fees.rs
```diff
@@ -0,0 +1,216 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use js_chain_libs::*;
+    use address::*;
+
+    const HLEN: usize = 4;
+    const SLEN: usize = 13;
+
+    // TODO: compare raw bytes now that we have complete test vectors for this
```
SebastienGllmt, line 120: This is no longer relevant, right? Since we know `Hash` is the right size

rust/src/fees.rs
```diff
@@ -0,0 +1,216 @@
...
+    }
+    fn bob_stake() -> Keyhash {
+        Keyhash::new(vec![6; HLEN])
+    }
+    fn bob_addr() -> Address {
+        BaseAddress::new(3, AddrCred::from_keyhash(bob_pay()), AddrCred::from_keyhash(bob_stake())).to_address()
+    }
+
+    fn make_witnesses_vkey(tx: &TransactionBody, pks: Vec<&PrivateKey>) -> TransactionWitnessSet {
+        // TODO: implement non-mock at some point?
```
SebastienGllmt, line 151: This is just pending the tx signing test vectors, right?

rust/src/fees.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    #[test]
+    fn tx_simple_utxo() {
+        let mut inputs = TransactionInputs::new();
+        inputs.add(TransactionInput::new(genesis_id(), 0));
+        let mut outputs = TransactionOutputs::new();
+        outputs.add(TransactionOutput::new(alice_addr(), 10));
+        let body = TransactionBody::new(inputs, outputs, 94, 10);
+        let w = make_witnesses_vkey(&body, vec![&alice_key()]);
+        //assert_eq!(body.to_bytes().len() + w.to_bytes().len(), 94);
```
SebastienGllmt, line 169: do we still need this assert?

rust/src/fees.rs
```diff
@@ -0,0 +1,216 @@
...
+        let mut body = TransactionBody::new(inputs, outputs, 94, 10);
+        let mut certs = DelegationCertificates::new();
+        certs.add(DelegationCertificate::new_key_reg(alice_pay()));
+        body.set_certs(certs);
+        let w = make_witnesses_vkey(&body, vec![&alice_key()]);
+        assert_eq!(txsize(&body, &w), 92);
+    }
+
+    // #[test]
+    // fn tx_delegate_stake() {
```
SebastienGllmt, line 204: Why is this test commented out?

rust/src/address.rs
rooooooooob: Within the context of the other CBOR serialization, we encode addresses as a CBOR byte string which has the Bytes major type length tag first. For to/from bytes I just made it the raw bytes (the ones you would use for the data portion of bech32 but no human readable portion).

rust/src/address.rs
rooooooooob: Done

rust/src/fees.rs
rooooooooob: The result was that they used mock crypto for their test vectors for tx fee sizes. I removed all the tx size stuff from this PR anyway. We'll do it in another with the new way.

rust/src/fees.rs
rooooooooob: Yeah but it doesn't matter anymore since they scrapped txsize

rust/src/fees.rs
rooooooooob: We didn't have updates supported in the cddl-codegen library at this point in time I think.

rust/src/fees.rs
rooooooooob: I was going to use it it to verify the general transaction structure against theres since it could be the same size but with some bad structure... Although I guess it would probably just lead to us realizing we're mixing definite/indefinite encodings somewhere and they wouldn't match byte-wise for that.

rust/src/fees.rs
rooooooooob: I believe so. We just were testing size at first so I didn't bother as well.

rust/src/fees.rs
rooooooooob: No

rust/src/fees.rs
rooooooooob: I don't think there's much of a downside. It'd be similar to that `Transaction` or signed transaction from js-chain-libs. We were just in a very early stage that also had changing CDDL specs so I never implemented library usability things like this.

PR #7, opened 2020-05-29
Reward addresses

rooooooooob:


rust/src/address.rs
```diff
@@ -121,25 +122,42 @@ impl Address {
...
+                buf.push(header);
+                buf.extend(reward.payment.to_bytes());
+            },
         }
         println!("to_bytes({:?}) = {:?}", self, buf);
         buf
     }
 
-    pub fn from_bytes(data: Vec<u8>) -> Result<Self, JsValue> {
+    pub fn from_bytes(data: Vec<u8>) -> Result<Address, JsValue> {
```
rooooooooob, line 137: This was independent but needed when calling from wasm. wasm_bindgen can't figure out the `Self` when exporting to wasm even though here `Self` resolves to `Address`.

rooooooooob:
Merging without test vectors - it is basically the same as the EnterpriseAddress format besides the header so there's not whole lot to mess up, and IOHK mentioned they were waiting before putting out any test vectors for reward addresses. We can add them later when the address format as a whole stabilizes and they release them, plus the library is marked experimental anyway.

PR #8, opened 2020-06-02
Fee calculation + partial shelley.cddl generation

rooooooooob:
This uses just a subset of shelley.cddl to generate that we'll need to
making transactions so we can calculate fees.

### DIFF
--- a/rust/src/address.rs
+++ b/rust/src/address.rs
@@ -206,9 +206,11 @@ impl Address {
         // bits 7-4: 1000
         let header = data[0];
         let network = header & 0x0F;
-        let hash_len = 28;
+        const HASH_LEN: usize = AddrKeyHash::BYTE_COUNT;
+        // should be static assert but it's maybe not worth importing a whole external crate for it now
+        assert_eq!(ScriptHash::BYTE_COUNT, HASH_LEN);
         let read_addr_cred = |bit: u8, pos: usize| {
-            let hash_bytes: [u8; 28] = data[pos..pos+hash_len].try_into().unwrap();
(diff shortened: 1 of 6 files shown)


rust/src/address.rs
```diff
@@ -29,54 +29,100 @@ fn variable_nat_encode(mut num: u64) -> Vec<u8> {
 }
 
 #[derive(Debug, Clone, Eq, Ord, PartialEq, PartialOrd)]
-enum AddrCredType {
-    Key(Keyhash),
-    Script(Scripthash),
+enum StakeCredType {
```
rooooooooob, line 32: this was basically `stake_credential` in `shelley.cddl` so it was renamed and (de)serialization was implemented so we could call it from the generated code.

rust/src/crypto.rs
```diff
@@ -327,4 +329,57 @@ macro_rules! impl_signature {
     };
 }
 
-impl_signature!(Ed25519Signature, Vec<u8>, crypto::Ed25519);
\ No newline at end of file
+impl_signature!(Ed25519Signature, Vec<u8>, crypto::Ed25519);
+
+macro_rules! impl_hash_type {
```
rooooooooob, line 334: From the cddl we'd just have all the hash types all as `Vec<u8>` which isn't the most helpful since they're always fixed-sized, so this macro is just to quickly add on whatever hash types we need and have them do size-checks/etc.

rooooooooob:
`prelude.rs`, `lib.rs` and `serialization.rs` are the files taken from the code generator. Only minor changes were made (renaming a few fields, changing `Coin` to `u64`) to those.

PR #12, opened 2020-07-01
Wasm-friendly interface change + cddl regen

rooooooooob:
1) Swap all wasm-exposed functions to take by-ref for rust-wasm structs
   and clone to be more foolproof memory-wise from js/wasm

2) Serialize everything using definite encoding instead of indefinite
   which a) should be more likely to be compatible with IOHK's haskell
   code and b) takes up 5% or so less space for txs

3) Some additions to shelley.cddl were made by IOHK since we last generated
    using it as well like reward accounts so those were added too.

4) Fixed up tx creation js test to use signing

5) from_bytes functions now auto-choose JsValue vs DeserializeError based on current compilation target (necessary because JsValue panics when used on non-wasm)

The ordering of the lib/serialization files is totally changes since
this time I used the shelley.cddl file mostly untouched, whereas last
time it was edited to get rid of out-of-order declarations e.g.
transaction being declared before transaction_body which it references,
but now cddl-codegen can handle that.

### DIFF
--- a/rust/pkg/cddl_lib.js.flow
+++ b/rust/pkg/cddl_lib.js.flow
@@ -72,6 +72,11 @@ declare export class Address {
    * @returns {Address}
    */
   static from_bech32(bech_str: string): Address;
+
+  /**
+   * @returns {number}
+   */
+  network_id(): number;
 }
 declare export class BaseAddress {
   free(): void;
@@ -962,7 +967,7 @@ declare export class PoolParams {
    * @param {BigInt} pledge
    * @param {BigInt} cost
    * @param {UnitInterval} margin
-   * @param {RewardAccount} reward_account
+   * @param {RewardAddress} reward_account
    * @param {AddrKeyHashes} pool_owners
    * @param {Relays} relays
    * @param {PoolMetadata | void} pool_metadata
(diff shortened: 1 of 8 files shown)


v-almonacid:
@rooooooooob shouldn't `pub fn from_bytes(bytes: Vec<u8>)` in `crypto::impl_hash_type` also take a `&[u8]`? 🤔 

https://github.com/Emurgo/cardano-serialization-lib/blob/9a7904e80f75a9b49dbfa2adb08eb3bb1e86e0b3/rust/src/crypto.rs#L647

rooooooooob:
> @rooooooooob shouldn't `pub fn from_bytes(bytes: Vec<u8>)` in `crypto::impl_hash_type` also take a `&[u8]`? thinking
> 
> https://github.com/Emurgo/cardano-serialization-lib/blob/9a7904e80f75a9b49dbfa2adb08eb3bb1e86e0b3/rust/src/crypto.rs#L647

No. This is meant to be exposed to js/wasm, hence the `JsValue` return type. For from-rust usage, if you have an exact size already there's a `From` trait impl for `[u8; BYTE_COUNT]` and I'm about to push a commit that will give you a `FromBytes` trait that is general-sized, and takes a `&[u8]`.

PR #17, opened 2020-07-08
Slight modifications to min_fee & coin

SebastienGllmt:
- `min_fee` now uses the `Coin` type (to match other places in the code)
- `min_fee` now takes the coefficient and constants as network parameters
- Adding some more utility functions to `Coin`

### DIFF
--- a/rust/src/fees.rs
+++ b/rust/src/fees.rs
@@ -26,7 +26,7 @@ pub fn min_fee(tx: &Transaction, coefficient: &Coin, constant: &Coin) -> Result<
         .ok_or(JsValue::from_str("Checked call failed"))
 }
 
-pub fn txsize(tx: &Transaction) -> usize {
+fn txsize(tx: &Transaction) -> usize {
     const UINT: usize = 5;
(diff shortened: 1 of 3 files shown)


rust/src/fees.rs
```diff
@@ -17,11 +17,13 @@ fn cbor_uint_length(x: u64) -> usize {
...
 #[wasm_bindgen]
-pub fn min_fee(tx: &Transaction) -> usize {
-    const A: usize = 500;
-    const B: usize = 2;
-    let size = txsize(tx);
-    A * size + B
+pub fn min_fee(tx: &Transaction, coefficient: &Coin, constant: &Coin) -> Result<Coin, JsValue> {
+    let size = fees::txsize(tx) as u64;
+    size
+        .checked_mul(coefficient.unwrap())
```
rooooooooob, line 23: Since we have those methods on `Coin` now we could probably just do `Coin::new(size).checked_mul(coefficient)?.checked_add(constant)` instead of having to unwrap them and map back and forth.

PR #20, opened 2020-07-10
Create tx builder

SebastienGllmt:
Background: in Yoroi, we need to display the user the expected fee of a transaction BEFORE they actually provide they key to sign the transaction.

This is not easy because the fee of the transaction depends on the witnesses (which are only generated once you sign). To solve this, we have a tx builder interface that can tell us what the fee will be as we're building the transaction

### DIFF
--- a/rust/src/tx_builder.rs
+++ b/rust/src/tx_builder.rs
@@ -37,7 +37,7 @@ fn witness_keys_for_cert(cert_enum: &Certificate, keys: &mut BTreeSet<AddrKeyHas
     }
 }
 
-fn calc_fee(tx_builder: &TransactionBuilder) -> Result<Coin, JsValue> {
+fn estimate_fee(tx_builder: &TransactionBuilder) -> Result<Coin, JsValue> {
     let body = tx_builder.build()?;
 
     let fake_key_root = Bip32PrivateKey::from_bip39_entropy(
@@ -220,14 +220,14 @@ impl TransactionBuilder {
     }
     pub fn get_fee_or_calc(&self) -> Result<Coin, JsValue> {
         match &self.fee {
-            None => self.calc_fee(),
+            None => self.estimate_fee(),
             Some(x) => Ok(x.clone()),
         }
     }
 
     pub fn add_change_if_needed(&mut self, address: &Address) -> Result<bool, JsValue> {
         let fee = match &self.fee {
-            None => self.calc_fee(),
+            None => self.estimate_fee(),
             // generating the change output involves changing the fee
             Some(_x) => return Err(JsValue::from_str("Cannot calculate change if fee was explicitly specified")),
         }?;
@@ -243,7 +243,7 @@ impl TransactionBuilder {
                     // this may over-estimate the fee by a few bytes but that's okay
                     amount: Coin::new(0x1_00_00_00_00),
                 });
-                let new_fee = copy.calc_fee()?;
+                let new_fee = copy.estimate_fee()?;
                 match input_total > output_total.checked_add(&new_fee)? {
                     false => return Ok(false), // not enough input to covert the extra fee from adding an output
                     true => {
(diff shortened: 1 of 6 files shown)


rust/src/tx_builder.rs
```diff
@@ -0,0 +1,164 @@
...
+            // if user explicitly specified a fee already, use that one
+            Some(_x) => {},
+            // otherwise, use the maximum fee possible
+            None => {
+                self.set_fee(&Coin::new(0x1_00_00_00_00));
+            },
+        };
+        let body = self.build()?;
+        let witness_set = TransactionWitnessSet {
+            vkeys: Some(Vkeywitnesses::new()),
```
SebastienGllmt, line 150: TODO: I need to make these actually be populated with fake witnesses. Need to spend time looking up how to generate fake Byron witnesses, etc.

rust/src/tx_builder.rs
```diff
@@ -0,0 +1,164 @@
...
+impl TransactionBuilder {
+    // We have to know what kind of inputs these are to know what kind of mock witnesses to create since
+    // 1) mock witnesses have different lengths depending on the type which changes the expecting fee
+    // 2) Witnesses are a set so we need to get rid of duplicates to avoid over-estimating the fee
+    pub fn add_key_input(&mut self, hash: &AddrKeyHash, input: &TransactionInput) {
+        self.inputs.add(input);
+        self.input_types.vkeys.insert(hash.clone());
+    }
+    pub fn add_script_input(&mut self, hash: &ScriptHash, input: &TransactionInput) {
+        self.inputs.add(input);
```
SebastienGllmt, line 69: I'm not really sure how script witnesses work and if this approach even works for script witnesses to be honest. We won't need them for the initial release though so maybe it's okay.

rust/src/tx_builder.rs
```diff
@@ -0,0 +1,164 @@
...
+        },
+        CertificateEnum::StakeDelegation(cert) => {
+            keys.insert(cert.stake_credential().to_keyhash().unwrap());
+        },
+        CertificateEnum::PoolRegistration(cert) => {
+            for owner in &cert.pool_params().pool_owners().0 {
+                keys.insert(owner.clone());
+            }
+            keys.insert(
+                AddrKeyHash::from_bytes(cert.pool_params().operator().to_bytes()).unwrap()
```
SebastienGllmt, line 42: In the 3 places in this function, we can safely convert these to `AddrKeyHash` because both `AddrKeyHash` and the other key hash types all have the same hash size and are generated by the same hashing algorithm.
Therefore, the hash matching => the underlying key matches

The code is a slightly ugly way to represent that though. Not sure if there's is a better way though.
One alternative is to add a generic From/Into between all the KeyHash types in `crypto.rs` (`GenesisDelegateHash`, `PoolKeyHash`, `AddrKeyHash`, `VRFKeyHash`). However, that doesn't really feel less hacky (but maybe more readable)

rooooooooob: Yes, in terms of size it doesn't matter since scripthash and addrkeyhash are 28 bytes, but the problem is that as-is that unwrap will panic for scripthash staking credentials as `StakingCredential::to_keyhash()` returns `None` for script credentials. I see you're referring to the other hashes, but for `StakingCredential` it's not a matter of ugly, it will just panic using `to_keyhash()`.

We could do a conversion you said and make some Hash28/Hash32 (or whatever you want to call them - just using those since the shelley.cddl uses them) and make this a `BTreeSet<Hash28>` so it doesn't seem like we're reusing AddrKeyHash for hashes of other keys just because it's the same size.

rust/src/tx_builder.rs
```diff
@@ -0,0 +1,164 @@
...
+        self.certs = Some(certs.clone());
+        for cert in &certs.0 {
+            witness_keys_for_cert(cert, &mut self.input_types.vkeys);
+        };
+    }
+
+    pub fn set_withdrawals(&mut self, withdrawals: &Withdrawals) {
+        self.withdrawals = Some(withdrawals.clone());
+        for (withdrawal, _coin) in &withdrawals.0 {
+            self.input_types.vkeys.insert(withdrawal.payment_cred().to_keyhash().unwrap().clone());
```
rooooooooob, line 95: Here too

rust/src/tx_builder.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    pub fn build(&self) -> Result<TransactionBody, JsValue> {
+        let fee = self.fee.ok_or_else(|| JsValue::from_str("Fee not specified"))?;
+        let ttl = self.ttl.ok_or_else(|| JsValue::from_str("ttl not specified"))?;
+        Ok(TransactionBody {
+            inputs: self.inputs.clone(),
+            outputs: self.outputs.clone(),
+            fee: fee,
+            ttl: ttl,
+            certs: None,
```
rooooooooob, line 128: Shouldn't this include certs/withdrawals?

PR #21, opened 2020-07-10
Describe bootstrap witness types

SebastienGllmt:
Still need to write code to actually generate these witness types.

### DIFF
--- a/rust/src/address.rs
+++ b/rust/src/address.rs
@@ -240,7 +240,7 @@ impl Address {
             const HASH_LEN: usize = AddrKeyHash::BYTE_COUNT;
             // should be static assert but it's maybe not worth importing a whole external crate for it now
             assert_eq!(ScriptHash::BYTE_COUNT, HASH_LEN);
-            // read nth /bit/ at byte position /pos/
+            // checks the /bit/ bit of the header for key vs scripthash then reads the credential starting at byte position /pos/
             let read_addr_cred = |bit: u8, pos: usize| {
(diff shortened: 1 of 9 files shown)


rust/src/crypto.rs
```diff
@@ -402,28 +414,47 @@ impl Deserialize for Vkeywitnesses {
...
-// TODO: custom-write the 3 byte objects and generally make this useable for more than deserialization
 #[wasm_bindgen]
 #[derive(Clone)]
 pub struct BootstrapWitness {
     vkey: Vkey,
     signature: Ed25519Signature,
-    index_2: Vec<u8>,
-    index_3: Vec<u8>,
-    index_4: Vec<u8>,
+    chain_code: Vec<u8>,
```
rooooooooob, line 422: I know it's a draft still, but are we going to make these separate structs or how/where will the validity checks be done? I assume they're not arbitrary bytes.

SebastienGllmt: I keep all of these as Vec<u8> since they're not easy to represent in CDDL

rust/src/address.rs
```diff
@@ -210,6 +236,7 @@ impl Address {
             const HASH_LEN: usize = AddrKeyHash::BYTE_COUNT;
             // should be static assert but it's maybe not worth importing a whole external crate for it now
             assert_eq!(ScriptHash::BYTE_COUNT, HASH_LEN);
+            // read nth /bit/ at byte position /pos/
```
rooooooooob, line 239:
```suggestion
            // checks the /bit/ bit of the header for key vs scripthash then reads the credential starting at byte position /pos/
```

PR #23, opened 2020-07-15
Clarify u64/u32/etc size limitations from the cddl uints

rooooooooob:
Metadata needs to be u64, the encoding of metadata contents needed u64
as well.

Epoch/slot/tx index/cert index are fine as u32 as it will be many
hundreds of thousands or more of years until that is an issue.

prelude.rs was also split up into error.rs and the rest into utils.rs
because half of it was either hand-written or hand-ededit from
cddl-codgen and the file itself is a static file within cddl-codgen so
there shouldn't be any changes as a result of re-generation of the cddl.


src/index.spec.ts
```diff
@@ -84,15 +84,15 @@ describe('Transactions', () => {
           CardanoWasm.Address.from_bytes(
             Buffer.from('61a6274badf4c9ca583df893a73139625ff4dc73aaa3082e67d6d5d08e', 'hex'),
           ),
-          // we can construct Coin from both BigInt (here) or from a string (below in fee)
-          CardanoWasm.Coin.new(BigInt(1)),
+          // we can construct BigNum (Coin) from both a js BigInt (here) or from a string (below in fee)
```
rooooooooob, line 87: The downside is that when you do `type Coin = BigNum` in rust while you can use it in rust directly like `Coin::new(0)` you can't do that in JS as it only exports the `BigNum` class and replaces all other things that type alias to `BigNum` as `BigNum` such as `Coin` or `TransactionMetadataLabel` and such.

If we want to allow direct usage from js using the aliases we could either copy the code for every u64-like type (lots of duplicate code) or make the entire class into a macro and invoke it a few times (which is kinda ugly imo since there's no real difference in the code) so I'm not sure which approach we want to take between those and the current case of it all being `BigNum` from the js side.

rust/src/utils.rs
```diff
@@ -1,73 +1,209 @@
...
       match raw.unsigned_integer() {
           Ok(value) => Ok(Self(value)),
-          Err(e) => Err(DeserializeError::new("Coin", DeserializeFailure::CBOR(e))),
+          Err(e) => Err(DeserializeError::new("BigNum", DeserializeFailure::CBOR(e))),
       }
   }
 }
 
+// Specifies an amount of ADA in terms of lovelace
+pub type Coin = BigNum;
```
SebastienGllmt, line 135: Is there a reason to keep `Coin` at all?

Issue #32, opened 2020-07-29
Allow different bech32 address prefixes


SebastienGllmt:
Currently the Address `to_bech32` function has a hard-coded prefix `addr`

However, it's entirely reasonable for people to want their own prefixes. Especially because for testnets, IOHK wants to use `addr_test` as the prefix for addresses

PR #43, opened 2020-08-09
feat: allow different bech32 address prefixes

AlanVerbner:
# Decription

This PR allows custom bech32 prefixes.

BREAKING CHANGE: Rust API requires an new `Option` parameter. JS bindings will accept `undefined` and default value will be used.

Closes: #32

# Questions

@SebastienGllmt in order to remove draft status there are two main things I need to finish / decide:

1. Do you like the `Option` approach or do you think it's best to add a new function? Something like:

```rust
pub fn to_bech32_with_prefix(&self, prefix: &str) -> String {
}
```
As far as I understand, it's breaking the Rust API but not the TS / Flow one. I'm not sure how this lib is used (we are using it from NodeJS) so it might be a better idea to keep the API stable. Let me know and I'll change it.

2. What do you think it's the best approach to test this? I can create a test vector using `cardano-cli` or try to generate one for each test you already have (I think you took them from [here](https://gist.github.com/KtorZ/b2e4e1459425a46df51c023fda9609c8))

### DIFF
--- a/rust/src/address.rs
+++ b/rust/src/address.rs
@@ -380,8 +380,8 @@ impl Address {
         })().map_err(|e| e.annotate("Address"))
     }
 
-    pub fn to_bech32(&self) -> String {
-        bech32::encode("addr", self.to_bytes().to_base32()).unwrap()
+    pub fn to_bech32(&self, prefix: Option<String>) -> String {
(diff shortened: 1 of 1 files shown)


v-almonacid:
> this matters for mobile clients since they have to build Rust bindings directly (can't run WASM in mobile devices)

I don't think this should be an issue.

PR #52, opened 2020-08-12
First draft for documentation

SebastienGllmt:
Not much included. Just setting everything up so we have a starting point


README.rst
```diff
@@ -0,0 +1,118 @@
+Cardano Serialization Lib
```
SebastienGllmt, line 1: I'm not sure if moving from a markdown to an RST is a requirement. I'm concerned this will affect our NPM package since NPM can parse markdown for a package, but maybe not RST

Ex: https://www.npmjs.com/package/@emurgo/cardano-serialization-lib-nodejs

doc/.sphinx/_templates/layout.html
```diff
@@ -0,0 +1,7 @@
+{% extends "!layout.html" %}
```
SebastienGllmt, line 1: I'm not sure this is used anywhere

README.rst
robcohen: Another way to do this, is to have it in markdown but have the toctrees parsed as RST -> https://recommonmark.readthedocs.io/en/latest/auto_structify.html#embed-restructuredtext

doc/.sphinx/_templates/layout.html
robcohen: It's not, just an example

README.rst
robcohen: The reason why I did it the way I did it was that I could use the include directive with RST files, so I could easily seperate the toctree from the README. Otherwise, github will render the toctree which is pretty ugly.

robcohen: I tried using the mdinclude directive, but it didn't work. if you can get it working that'd be dope.

robcohen:
Should I merge this?

PR #78, opened 2020-09-01
Switch table-types from BTreeMap to LinkedHashMap

rooooooooob:
To keep the order between serialization/deserialization consisetent
instead of storing in a BTreeMap which would ignore the
deserialize/insertion order and always serialize/iterate in lex order,
this is switched to linked_hash_map::LinkedHashMap which preserves this
order.

This issue still exists for non-iterable map types (e.g. map structs)
but that would be harder to fix (manually keeping track of the order)
and is less important since we don't iterate struct fields outside of
serialization.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -90,6 +90,7 @@ dependencies = [
  "ed25519-dalek",
  "hex",
  "js-sys",
+ "linked-hash-map",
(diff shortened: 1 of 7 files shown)


SebastienGllmt:
Shouldn't this also require a PR to cddl-codegen?

PR #80, opened 2020-09-02
Remove rust nightly requirements

rooooooooob:
Due to lack of support on some build targets we will remove the nightly
requirement by just temporarily, until >32 size auto-trait impls for
arrays are implemented by ~October, manually writing out the KESSignature
struct but using heap-allocation.

### DIFF
--- a/rust/src/crypto.rs
+++ b/rust/src/crypto.rs
@@ -761,7 +761,53 @@ impl_hash_type!(BlockHash, 32);
 impl_hash_type!(VRFVKey, 32);
 impl_hash_type!(KESVKey, 32);
 // same for this signature
-impl_hash_type!(KESSignature, 448);
+//impl_hash_type!(KESSignature, 448);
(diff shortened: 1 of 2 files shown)


rust/src/crypto.rs
```diff
@@ -761,7 +761,53 @@ impl_hash_type!(BlockHash, 32);
 impl_hash_type!(VRFVKey, 32);
 impl_hash_type!(KESVKey, 32);
 // same for this signature
-impl_hash_type!(KESSignature, 448);
+//impl_hash_type!(KESSignature, 448);
+// TODO: when >32 size trait implementations are out of nightly and into stable
```
v-almonacid, line 765: Note that this can take some time, since even in the current stable some tier-3 components are still missing is targets `armv7-apple-ios` and `i386-apple-ios`

SebastienGllmt (approved):

PR #84, opened 2020-09-17
(breaking change) Fix ByronAddress generation from key

SebastienGllmt:
The `from_icarus_key` functions that going forward, any new network for Cardano would have `protocolMagic == networkId` since it would make a lot of things simpler. Unfortunately, it seems it's not the case (the new Shelley testnet uses `1097911063` as the protocol magic). I have no idea why IOHK did this.

Correspondingly, I change `from_icarus_key` to accepted a protocol magic instead of a network id. Since both of these are stored as simple numbers, I changed the function name to ensure that anybody upgrading our library properly gets a a compile-time error instead of the wrong addresses generated.

This PR also adds a new `NetworkInfo` struct just to make things cleaner.

### DIFF
--- a/rust/src/address.rs
+++ b/rust/src/address.rs
@@ -189,7 +189,7 @@ impl ByronAddress {
     /// returns the byron protocol magic embedded in the address, or mainnet id if none is present
     /// note: for bech32 addresses, you need to use network_id instead
     pub fn byron_protocol_magic(&self) -> u32 {
-        match self.0.attributes.network_magic {
+        match self.0.attributes.protocol_magic {
             Some(x) => x,
             None => NetworkInfo::mainnet().protocol_magic(), // mainnet is implied if omitted
         }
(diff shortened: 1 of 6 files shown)


rust/src/address.rs
```diff
@@ -165,21 +200,22 @@ impl ByronAddress {
...
         // recall: in Byron mainnet, the network_id is omitted from the address to save a few bytes
-        let mainnet_network_id = 764824073;
         // so here we return the mainnet id if none is found in the address
 
-        match self.0.attributes.network_magic {
-            // although mainnet should never be explicitly added, we check for it just in case
-            Some(x) => if x == mainnet_network_id { 0b0001 } else { 0b000 },
-            None => 0b0001, // mainnet is implied if omitted
+        // mainnet is implied if omitted
+        let protocol_magic = self.0.attributes.network_magic.unwrap_or(NetworkInfo::mainnet().protocol_magic());
```
rooooooooob, line 214: What if we just used `let protocol_magic = self.0byron_protocol_magic()` here since that logic is already done above?

rust/src/address.rs
```diff
@@ -154,7 +189,7 @@ impl ByronAddress {
     /// returns the byron protocol magic embedded in the address, or mainnet id if none is present
     /// note: for bech32 addresses, you need to use network_id instead
     pub fn byron_protocol_magic(&self) -> u32 {
-        let mainnet_network_id = 764824073;
+        let mainnet_network_id = NetworkInfo::mainnet().protocol_magic();
```
rooooooooob, line 192: It's probably fine as-is as it's isolated in the implementation of the function that only the lib's developers will see, but maybe it'd be better to not call it network_id since we now have the distinction between network id and network/protocol magic?

rooooooooob (commented):
In rust it's fine for accidentally mixing up network id / protocol magic as the type system catches it, but from JS they're both just numbers, but then when going through the wasm boundary I'm pretty sure network id would get truncated modulo 256 silently. Do we care to prevent mixing them up by accident? I guess any tx just simply wouldn't be accepted by the network right? So maybe that's totally fine. If it weren't for how clunky wasm is with memory-freeing a wrapper struct would be nice to strongly type them, but using it without memory leaks from wasm would make this ugly.

rooooooooob (approved):
LGTM

PR #95, opened 2020-11-13
Asmjs build

dractw:
Downgrading wasm-bindgen due to wasm-opt bug with `Exported global cannot be mutable`
https://github.com/rustwasm/wasm-pack/issues/886

asmjs build runs successfully


SebastienGllmt (approved):
Thanks a lot for this!

PR #97, opened 2020-11-18
Add size checks to pool registration cert objects

rooooooooob:
Add in checks for: Ipv4, Ipv6, Url (Moved to dedicated struct)

Split DnsName (string alias) into DNSRecordAorAAAA / DNSRecordSRV to be
more descriptive for the two use-cases and to allow us to potentially
add in types specific to each type of DNS record during creation.

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -827,31 +827,126 @@ type Port = u16;
 
 #[wasm_bindgen]
 #[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd)]
-pub struct Ipv4(Vec<u8>);
+pub struct Ipv4([u8; 4]);
 
 to_from_bytes!(Ipv4);
 
 #[wasm_bindgen]
 impl Ipv4 {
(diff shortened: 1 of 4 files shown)


rust/src/lib.rs
```diff
@@ -827,31 +827,126 @@ type Port = u16;
...
+        Self::new_impl(dns_name).map_err(|e| JsError::from_str(&e.to_string()))
+    }
+
+    pub (crate) fn new_impl(dns_name: String) -> Result<DNSRecordSRV, DeserializeError> {
+        if dns_name.len() <= 64 {
+            Ok(Self(dns_name))
+        } else {
+            Err(DeserializeError::new("DNSRecordSRV", DeserializeFailure::OutOfRange{
+                min: 0,
+                max: 64,
```
SebastienGllmt, line 944: Since this `64` appears in multiple places, maybe we should make it a constant somewhere? I know a lot of pool operators complain about the 64-byte limit so I can imagine it being increased in the future.

SebastienGllmt (approved):

PR #98, opened 2020-11-27
Add support for the Allegra hardfork

rooooooooob:
Users will not have to change API for non-metadata TX creation, but however will for ones with metadata.

The metadata API change changed from:
```rust
let mut old_metadata = TransactionMetadata::new();
old_metadata.insert(/*label*/, /*metadatum*/);
```
to
```rust
let mut old_metadata = GeneralTransactionMetadata::new();
old_metadata.insert(/*label*/, /*metadatum*/);
let mut new_metadata = TransactionMetadata::new(&old_metadata);

// possibly add in Allegra native script metadata:
let mut scripts = NativeScripts::new();
scripts.add(/*create script here*/);
new_metadata.set_native_scripts(&scripts);
```

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -208,11 +208,12 @@ pub struct TransactionBody {
     inputs: TransactionInputs,
     outputs: TransactionOutputs,
     fee: Coin,
-    ttl: u32,
+    ttl: Option<Slot>,
     certs: Option<Certificates>,
     withdrawals: Option<Withdrawals>,
     update: Option<Update>,
     metadata_hash: Option<MetadataHash>,
+    validity_start_interval: Option<Slot>,
 }
 
 to_from_bytes!(TransactionBody);
@@ -231,8 +232,8 @@ impl TransactionBody {
         self.fee.clone()
     }
 
-    pub fn ttl(&self) -> u32 {
-        self.ttl.clone()
+    pub fn ttl(&self) -> Option<Slot> {
+        self.ttl
     }
 
     pub fn set_certs(&mut self, certs: &Certificates) {
@@ -266,21 +267,30 @@ impl TransactionBody {
         self.metadata_hash.clone()
     }
 
+    pub fn set_validity_start_interval(&mut self, validity_start_interval: Slot) {
+        self.validity_start_interval = Some(validity_start_interval)
+    }
+
+    pub fn validity_start_interval(&self) -> Option<Slot> {
+        self.validity_start_interval
+    }
+
     pub fn new(
         inputs: &TransactionInputs,
         outputs: &TransactionOutputs,
         fee: &Coin,
-        ttl: u32,
+        ttl: Slot,
     ) -> Self {
         Self {
             inputs: inputs.clone(),
             outputs: outputs.clone(),
             fee: fee.clone(),
-            ttl: ttl,
+            ttl: Some(ttl),
             certs: None,
             withdrawals: None,
             update: None,
             metadata_hash: None,
+            validity_start_interval: None,
         }
     }
 }
@@ -1111,35 +1121,10 @@ impl Withdrawals {
 }
 
 #[wasm_bindgen]
-#[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd)]
-pub struct MultisigScripts(Vec<MultisigScript>);
-
-to_from_bytes!(MultisigScripts);
-
-#[wasm_bindgen]
-impl MultisigScripts {
-    pub fn new() -> Self {
-        Self(Vec::new())
-    }
-
-    pub fn len(&self) -> usize {
-        self.0.len()
-    }
-
-    pub fn get(&self, index: usize) -> MultisigScript {
-        self.0[index].clone()
-    }
-
-    pub fn add(&mut self, elem: &MultisigScript) {
-        self.0.push(elem.clone());
-    }
-}
-
-#[wasm_bindgen]
 #[derive(Clone)]
 pub struct TransactionWitnessSet {
     vkeys: Option<Vkeywitnesses>,
-    scripts: Option<MultisigScripts>,
+    scripts: Option<NativeScripts>,
     bootstraps: Option<BootstrapWitnesses>,
 }
 
@@ -1155,11 +1140,11 @@ impl TransactionWitnessSet {
         self.vkeys.clone()
     }
 
-    pub fn set_scripts(&mut self, scripts: &MultisigScripts) {
+    pub fn set_scripts(&mut self, scripts: &NativeScripts) {
         self.scripts = Some(scripts.clone())
     }
 
-    pub fn scripts(&self) -> Option<MultisigScripts> {
+    pub fn scripts(&self) -> Option<NativeScripts> {
         self.scripts.clone()
     }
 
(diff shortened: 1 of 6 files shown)


rust/src/lib.rs
```diff
@@ -208,11 +208,12 @@ pub struct TransactionBody {
     inputs: TransactionInputs,
     outputs: TransactionOutputs,
     fee: Coin,
-    ttl: u32,
+    ttl: Option<Slot>,
```
rooooooooob, line 211: Are TTL and validity start intervals the only places we use absolute slots and the others are relative or how is it? We should probably have two type aliases here to be clear even if they're stripped out in the wasm builds.

rust/src/lib.rs
```diff
@@ -266,21 +267,30 @@ impl TransactionBody {
...
+    pub fn validity_start_interval(&self) -> Option<Slot> {
+        self.validity_start_interval
+    }
+
     pub fn new(
         inputs: &TransactionInputs,
         outputs: &TransactionOutputs,
         fee: &Coin,
-        ttl: u32,
+        ttl: Slot,
```
rooooooooob, line 282: I left this field as required in the constructor since:
1) you mentioned it's a good idea for users to put it in
2) the API doesn't change

I can take it out if you want

rust/src/lib.rs
```diff
@@ -1182,17 +1167,18 @@ impl TransactionWitnessSet {
 
 #[wasm_bindgen]
 #[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd)]
-pub struct MsigPubkey {
+pub struct ScriptPubkey {
```
rooooooooob, line 1170: I left these names as the new changes ones even though they're basically the same thing since
1) I don't think anyone is using any of the multisig things anyway
2) it's now much more general than multisig so it's less confusing this way

SebastienGllmt (approved):
LGTM
Do we want to merge this to master though? Maybe it's better to use a separate branch until the hardfork actually happens (although maybe that would make it harder to then do the Mary changes)

rust/src/metadata.rs
```diff
@@ -247,16 +248,14 @@ impl TransactionMetadatumLabels {
...
-    linked_hash_map::LinkedHashMap<TransactionMetadatumLabel, TransactionMetadatum>,
-);
+pub struct GeneralTransactionMetadata(LinkedHashMap<TransactionMetadatumLabel, TransactionMetadatum>);
 
-to_from_bytes!(TransactionMetadata);
+to_from_bytes!(GeneralTransactionMetadata);
 
 #[wasm_bindgen]
-impl TransactionMetadata {
+impl GeneralTransactionMetadata {
```
rooooooooob, line 256: I thought about having this stay in the same struct as before and simply adding a `native_scripts()` and `set_native_scripts()` field to it to make it so the API stays the same for anyone not using the new native scripts metadata but then the `insert()` , `len()` and `get()` might be confusing whether they refer to the old metadata instead of indexing/getting the length of the native script portion. We could do that anyway if we want to not change the API. I didn't use the direct code-gen since that would've been even more structs generated (eg `TransactionMetadataEnum`, `TransactionMetadataKind`, and `AllegraMetadata` with more indirection and it being treated like the other enums) but since one is basically a subset of the other I combined them somewhat.

rust/src/metadata.rs
```diff
@@ -717,11 +747,60 @@ impl Deserialize for TransactionMetadata {
...
             Ok(())
-        })().map_err(|e| e.annotate("TransactionMetadata"))?;
+        })().map_err(|e| e.annotate("GeneralTransactionMetadata"))?;
         Ok(Self(table))
     }
 }
 
+impl cbor_event::se::Serialize for TransactionMetadata {
+    fn serialize<'se, W: Write>(&self, serializer: &'se mut Serializer<W>) -> cbor_event::Result<&'se mut Serializer<W>> {
+        match &self.native_scripts() {
```
rooooooooob, line 757: There's a unit test at the bottom to test that the serialization is still fine, and the inner parts of these matches are from the codegen anyway.

rust/src/tx_builder.rs
```diff
@@ -389,19 +393,19 @@ impl TransactionBuilder {
 
     pub fn build(&self) -> Result<TransactionBody, JsError> {
         let fee = self.fee.ok_or_else(|| JsError::from_str("Fee not specified"))?;
-        let ttl = self.ttl.ok_or_else(|| JsError::from_str("ttl not specified"))?;
```
rooooooooob, line 392: Since it's no longer mandatory I didn't put it here even though the field is still required if you created it through the `TransactionBody::new()` field. I suppose we could make it so that either this is mandatory here or change the constructor to be optional?

rust/src/utils.rs
```diff
@@ -231,6 +231,49 @@ pub trait DeserializeEmbeddedGroup {
     ) -> Result<Self, DeserializeError> where Self: Sized;
 }
 
+pub struct CBORReadLen {
```
rooooooooob, line 234: cddl-codegen was updated to check array/map lengths during deserializaiton so this stuff was added.

rust/src/serialization.rs
```diff
@@ -1726,147 +1715,153 @@ impl DeserializeEmbeddedGroup for MsigPubkey {
...
             ret
-        })().map_err(|e| e.annotate("MsigAny"))
+        })().map_err(|e| e.annotate("ScriptAny"))
     }
 }
 
-impl DeserializeEmbeddedGroup for MsigAny {
-    fn deserialize_as_embedded_group<R: BufRead + Seek>(raw: &mut Deserializer<R>, len: cbor_event::Len) -> Result<Self, DeserializeError> {
+impl DeserializeEmbeddedGroup for ScriptAny {
+    fn deserialize_as_embedded_group<R: BufRead + Seek>(raw: &mut Deserializer<R>, /*/*read_len: &mut CBORReadLen, */*/len: cbor_event::Len) -> Result<Self, DeserializeError> {
```
rooooooooob, line 1811: The `DeserializeEmbeddedGroup` trait from before and now (after deserialize length checks on array/maps) are incompatible so I just commented out those parts here (and in other parts in this file) that interfered with it. If we re-generate the other parts of the library with those checks/other changes we could uncomment them out.

The newer code should generate less trait impls (only does DeserializeEmbeddedGroup when necessary - not often, instead of always) as well but I'm not sure how that translates to wasm size during packing, since there's also going to be the additional length checks.

rust/src/lib.rs
SebastienGllmt: I believe Timelock Start and Timelock Expire are both absolute slots as well. Additionally, `HeaderBody` also has this.

rust/src/lib.rs
SebastienGllmt: Yeah personally I can't think of a reason why somebody would ever not want to specify a TTL. However, it may be better to make it optional to properly mirror the spec. I'm okay with either options

rust/src/metadata.rs
SebastienGllmt: Yeah, I think this makes sense. I'm slightly concerned that the fact we're keeping the `TransactionMetadata` name (but not referring to something different) may cause subtle bugs for people though.

rust/src/tx_builder.rs
SebastienGllmt: Yeah I think removing it is fine.

PR #106, opened 2021-01-08
[WIP] Mary Support

rooooooooob:


rooooooooob:
All the core structs have been updated. The auxillary functionality still needs updating, e.g. tx builder, etc and whatever logic that might concern multiasset support beyond the data representation.

PR #108, opened 2021-01-08
Mary support

enzoc4:
continues #106 as I can't push there.

I didn't really make any logic changes, besides throwing an error when trying to get the minimum utxo value and the outputs include native assets. There are more details in the commit message.

### DIFF
--- a/rust/src/tx_builder.rs
+++ b/rust/src/tx_builder.rs
@@ -108,7 +108,7 @@ struct MockWitnessSet {
 #[derive(Clone, Debug)]
 struct TxBuilderInput {
     input: TransactionInput,
-    amount: Coin, // we need to keep track of the amount in the inputs for input selection
+    amount: Value, // we need to keep track of the amount in the inputs for input selection
 }
 
 #[wasm_bindgen]
@@ -135,29 +135,29 @@ impl TransactionBuilder {
     // We have to know what kind of inputs these are to know what kind of mock witnesses to create since
     // 1) mock witnesses have different lengths depending on the type which changes the expecting fee
     // 2) Witnesses are a set so we need to get rid of duplicates to avoid over-estimating the fee
-    pub fn add_key_input(&mut self, hash: &Ed25519KeyHash, input: &TransactionInput, amount: &Coin) {
+    pub fn add_key_input(&mut self, hash: &Ed25519KeyHash, input: &TransactionInput, amount: &Value) {
         self.inputs.push(TxBuilderInput {
             input: input.clone(),
             amount: amount.clone(),
         });
         self.input_types.vkeys.insert(hash.clone());
     }
-    pub fn add_script_input(&mut self, hash: &ScriptHash, input: &TransactionInput, amount: &Coin) {
+    pub fn add_script_input(&mut self, hash: &ScriptHash, input: &TransactionInput, amount: &Value) {
         self.inputs.push(TxBuilderInput {
             input: input.clone(),
             amount: amount.clone(),
         });
         self.input_types.scripts.insert(hash.clone());
     }
-    pub fn add_bootstrap_input(&mut self, hash: &ByronAddress, input: &TransactionInput, amount: &Coin) {
+    pub fn add_bootstrap_input(&mut self, hash: &ByronAddress, input: &TransactionInput, amount: &Value) {
         self.inputs.push(TxBuilderInput {
             input: input.clone(),
             amount: amount.clone(),
         });
         self.input_types.bootstraps.insert(hash.to_bytes());
     }
-
-    pub fn add_input(&mut self, address: &Address, input: &TransactionInput, amount: &Coin) {
+    
+    pub fn add_input(&mut self, address: &Address, input: &TransactionInput, amount: &Value) {
         match &BaseAddress::from_address(address) {
             Some(addr) => {
                 match &addr.payment_cred().to_keyhash() {
@@ -206,7 +206,7 @@ impl TransactionBuilder {
     }
 
     /// calculates how much the fee would increase if you added a given output
-    pub fn fee_for_input(&mut self, address: &Address, input: &TransactionInput, amount: &Coin) -> Result<Coin, JsError> {
+    pub fn fee_for_input(&mut self, address: &Address, input: &TransactionInput, amount: &Value) -> Result<Coin, JsError> {
         let mut self_copy = self.clone();
 
         // we need some value for these for it to be a a valid transaction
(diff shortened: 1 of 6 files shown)


rust/src/tx_builder.rs
```diff
@@ -367,11 +369,15 @@ impl TransactionBuilder {
                     address: address.clone(),
                     // maximum possible output to maximize fee from adding this output
                     // this may over-estimate the fee by a few bytes but that's okay
-                    amount: to_bignum(0x1_00_00_00_00),
+                    amount: Value::new(to_bignum(0x1_00_00_00_00)),
```
SebastienGllmt, line 372: Unfortunately this isn't sufficient because the change also needs to contain the difference between assets in the output and the assets in the input (since we need to send any leftover assets back as change)

rust/src/tx_builder.rs
```diff
@@ -127,6 +126,8 @@ pub struct TransactionBuilder {
     metadata: Option<TransactionMetadata>,
     validity_start_interval: Option<Slot>,
     input_types: MockWitnessSet,
+    mint: Option<Mint>,
+    constant_min_utxo_value: BigNum,
```
SebastienGllmt, line 130: Probably a better name is something like `ada_per_unit_size`

rust/src/tx_builder.rs
```diff
@@ -417,6 +428,21 @@ impl TransactionBuilder {
...
+            .outputs
+            .0
+            .iter()
+            .any(|output| output.amount().multiasset().is_some())
+        {
+            Err(JsError::from_str(
+                "Transaction with non-ada outputs is not implemented",
+            ))
+        } else {
+            Ok(self.constant_min_utxo_value)
```
SebastienGllmt, line 443: I would change this to just always return the mainnet min utxo value and add a comment that this will be replaced by formula once that formula is finalized.

enzoc4: I'm not sure I understand, do you want to

a) remove the parameter in the txbuilder constructor and always return 1 (a few tests need to be fixed in that case,as they rely on it)
or
b) don't throw an error and just return 1 or 2. And also leave the comment about the formula.

rust/src/tx_builder.rs
enzoc4: good point. I implemented that, but it doesn't make sense if I don't remove the error that's thrown when computing the min utxo value and there are native assets present.

also, not that it matters for this, but wouldn't `input_total - output_total` be a better upper bound for the fee estimation? After all, the change value would always be lower than that.

rust/src/tx_builder.rs
enzoc4: I wonder though, why can't we change the txbuilder to take the 3 constants of the formula in the spec as arguments and just compute the thing from there? Even if we don't have them right now, they are network parameters anyway.

rust/src/tx_builder.rs
```diff
@@ -304,18 +305,17 @@ impl TransactionBuilder {
...
-            .try_fold(
-                to_bignum(0),
-                |acc, ref tx_builder_input| acc.checked_add(&tx_builder_input.amount)
-            )
+            .try_fold(to_bignum(0), |acc, ref tx_builder_input| {
+                acc.checked_add(&tx_builder_input.amount.coin())
+            })
     }
     /// withdrawals and refunds
     pub fn get_implicit_input(&self) -> Result<Coin, JsError> {
```
SebastienGllmt, line 321: Probably `get_explicit_input`, `get_implicit_input` and `get_explicit_output` should return `Value` instead of `Coin`

rust/src/tx_builder.rs
SebastienGllmt: > wouldn't input_total - output_total be a better upper bound for the fee estimation

I think you're right as long as the `input_total` and `output_total` also take into account deposits, withdrawals, etc. A better estimation would reduce the transaction size by a single-digit number of bytes which is nice but also not super critical either. Up to you if you want to do this, or we can just create an open issue and improve this in the future.

rust/src/tx_builder.rs
SebastienGllmt: The thing is there are actually four parameters:

1) 3 parameters to calculate the approximate size of the transaction (outputSizeConstants in the spec)
2) One parameter to decide how many lovelaces each unit of this virtual size is worth (adaPerUTxOByte in the spec)

We could make the transaction builder take all four of these as parameters. I was thinking we could maybe just hardcode the three parameters of (1) and only take (2) as an argument, but if we think these will change based on the network maybe you're right we maybe should just create a struct to represent all of these and pass them into the tx builder. It would maybe also probably make it easier to port all the tests

I haven't been told what the values of the three parameters in (1) will be, only that the result of applying them will give 1 ADA when only ADA is used in the tx.

enzoc4: I see, I missed the part about (2). That's what `ada_per_unit_size` means, then. And right now (1) is hardcoded to (1, 0, 0), essentially, although it just throws an error, so it is more like (1, `undefined`, `undefined`).
 
I don't mind hardcoding (1) too with the mainnet min utxo value if you want (although I don't know it), I just think it is more consistent to leave as it is, because it was already parametric.

Tbh I think just taking (1) for now it's fine but I don't know that well the how's this library used and how hard it is to make a breaking change later.

rust/src/tx_builder.rs
enzoc4: I mean, I don't think it's important, I was mostly wondering if I was missing something.
If I implement arithmetic (`checked_add` and `checked_sub`) on `Value` accounting for multiassets too, then I can just use `input_total.checked_sub(output_total)` for the estimation, so I don't mind doing it too.

rust/src/tx_builder.rs
enzoc4: Hmmm, I guess I can add `checked_add` to `Value`, maybe `Eq` and `PartialOrd` too (and expose them to js?). 
With `checked_sub` the code for the `add_change_if_needed` would look a bit simpler (or at least, the same as it was before with only scalar values), so that's worth adding too.

rust/src/tx_builder.rs
SebastienGllmt: Okay, what we have right now is a total hack, but adding these 4 parameters as an input to the TxBuilder requires refactoring things so if you want to just leave it as-is for now I'm okay with it (change `ada_per_unit_size` to a struct in a follow-up PR)

SebastienGllmt:
FYI this PR may affect us: https://github.com/input-output-hk/cardano-ledger-specs/pull/2092

I think for us it's fine because we defined MintAssets as `i128` so we're safe regardless

SebastienGllmt (approved):

rhyslbw:
Before this is merged @SebastienGllmt  @enzoc4 , can we consider using  the domain term `TokenBundle` instead of  `MultiAsset` ? We don't want to diverge from documentation and the spec

SebastienGllmt:
@rhyslbw the term MultiAsset comes from the binary specification. Our serialization code is automatically generated from the binary spec so it automatically uses the same name. If you want to change the name, you'll have to change the binary spec.

rhyslbw:
@SebastienGllmt ok thanks for confirming

PR #142, opened 2021-04-17
Add namespace to script hashing

SebastienGllmt:
I got the chance to look more into the issue of the script hashing requiring a prefix, and I think we should the function based on two things

1. After looking, it looks like the raw hash isn't used anywhere. The prefix is always used. Exposing a regular `hash` function is taking a risk because people using our library instead and then possibly accidentally lock funds forever because of it. I'd rather not take the risk for something unused
2. It looks like there will be more prefixes used in the future -- basically one for every language. Therefore, we should make this a function parameter instead of hardcoding the value `0`.


ashisherc (approved):
👍

PR #154, opened 2021-05-14
Fixed build on stable rust

ypopovych:
Hi.

I fixed the build on the current stable Rust. Old cryptoxide library can't be built on new stable.

I have run tests, but only on desktop (without wasm).

Check it, please. And if everything is ok, can you, please, release a new version. We need it for Swift wrapper.

Thanks,
Yehor @ Tesseract

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -17,9 +17,9 @@ checksum = "1d49d90015b3c36167a20fe2810c5cd875ad504b39cff3d4eae7977e6b7c1cb2"
 
 [[package]]
 name = "bech32"
-version = "0.7.2"
+version = "0.7.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cdcf67bb7ba7797a081cd19009948ab533af7c355d5caf1d08c777582d351e9c"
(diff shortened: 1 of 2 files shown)


thinkocapo:
Thanks @ypopovych this fixed it for me, otherwise I would get an error on compiling `cryptoxide` that looked like:

```
   Compiling rand_core v0.3.1
error[E0518]: attribute should be applied to function or closure
  --> /Users/<user>/.cargo/registry/src/github.com-1ecc6299db9ec823/cryptoxide-0.2.1/src/chacha/sse2.rs:38:1
   |
38 | #[inline]
   | ^^^^^^^^^ not a function or closure

   Compiling generic-array v0.14.4
   Compiling block-padding v0.1.5
   Compiling ed25519 v1.0.3
   Compiling rand_chacha v0.1.1
   Compiling clear_on_drop v0.2.4
error: aborting due to previous error

For more information about this error, try `rustc --explain E0518`.
error: could not compile `cryptoxide`
```

gszabo:
Hi! When is this going to be released to crates.io? I have the same compiler problem when I try to use the latest (7.0.0) crate.

SebastienGllmt:
@gszabo just pushed this as part of 7.1.0

PR #160, opened 2021-06-01
Alonzo initial work

rooooooooob:
[ X ] Plutus components generation (initial generation + clean-up / some refactoring)
[ X ] `ConstrPlutusData` custom impl
[ X ] re-generate existing portions of code that made sense to be codegenned again
[ X ] re-write `TransactionMetadata` -> `AuxilaryData` to handle new alonzo format/plutus
[ X ] re-write `TransactioOutput` (codegen can't handle it)
[ X ] re-write `MoveInstantaneousReward`
[ X ] Updates to existing portions of library

### DIFF
--- a/rust/src/tx_builder.rs
+++ b/rust/src/tx_builder.rs
@@ -91,7 +91,7 @@ fn min_fee(tx_builder: &TransactionBuilder) -> Result<Coin, JsError> {
     let full_tx = Transaction {
         body,
         witness_set,
-        metadata: tx_builder.metadata.clone(),
+        auxiliary_data: tx_builder.auxiliary_data.clone(),
     };
     fees::min_fee(&full_tx, &tx_builder.fee_algo)
 }
@@ -124,7 +124,7 @@ pub struct TransactionBuilder {
     ttl: Option<Slot>, // absolute slot number
     certs: Option<Certificates>,
     withdrawals: Option<Withdrawals>,
-    metadata: Option<TransactionMetadata>,
+    auxiliary_data: Option<AuxiliaryData>,
     validity_start_interval: Option<Slot>,
     input_types: MockWitnessSet,
     mint: Option<Mint>,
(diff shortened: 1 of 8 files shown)


SebastienGllmt:
https://github.com/input-output-hk/cardano-ledger-specs/pull/2355/files this PR makes changes to the CDDL spec

rust/src/tx_builder.rs
```diff
@@ -455,12 +462,17 @@ impl TransactionBuilder {
...
-                Some(x) => Some(utils::hash_metadata(x)),
+                Some(x) => Some(utils::hash_auxiliary_data(x)),
             },
             validity_start_interval: self.validity_start_interval,
             mint: self.mint.clone(),
+            // TODO: update for use with Alonzo
+            script_data_hash: None,
+            collateral: None,
+            required_signers: None,
+            network_id: None,
```
vantuz-subhuman, line 475: @rooooooooob , there must be setters added for these new fields, right?

yushih (commented):
I think this line https://github.com/Emurgo/cardano-serialization-lib/blob/1a1561df4104f11f44446e5bb9b5a2415b5a7ecd/rust/src/lib.rs#L743 should be changed to `MoveInstantaneousRewardsCert(MoveInstantaneousReward)`.

rust/src/tx_builder.rs
rooooooooob: Maybe, I just updated the bare minimum in the tx builder as those are all optional fields. You can still manually set them by manual building tx. I'm not sure about the tx creation process for all alonzo-era txs so I left it like this.

rooooooooob:
There's been another update regarding some of the bigint types after a discussion with Jared that was made to the CDDL today but I will address that in a separate PR later as I will be providing more usable wrappers around those anyway.

vantuz-subhuman (approved):

PR #165, opened 2021-06-07
Added unspent output data structure

alessandrokonrad:
The proposed Dapp connector https://github.com/cardano-foundation/CIPs/pull/88/files
requires an unspent output data structure in cbor.

### DIFF
--- a/rust/src/utils.rs
+++ b/rust/src/utils.rs
@@ -68,19 +68,17 @@ macro_rules! to_from_bytes {
 #[derive(Clone, Debug)]
 pub struct TransactionUnspentOutput {
     input: TransactionInput,
-    value: Value,
-    address: Address
+    output: TransactionOutput
 }
 
 to_from_bytes!(TransactionUnspentOutput);
 
 #[wasm_bindgen]
 impl TransactionUnspentOutput {
-    pub fn new(input: &TransactionInput, value: &Value, address: &Address) -> TransactionUnspentOutput {
+    pub fn new(input: &TransactionInput, output: &TransactionOutput) -> TransactionUnspentOutput {
         Self {
             input: input.clone(),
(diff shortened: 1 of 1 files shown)


rooooooooob:
@alessandrokonrad  Ah yes, thanks for pointing this out.  This was an oversight when I was drafting the first version of the spec. It was based on the ergo one which the regular output was enough to create an input out of. The CIP spec is not joined with this library directly so we're going to need to specify a way in that spec first that is implementation-independent. The spec so far directly refers to the [shelley-ma.cddl](https://github.com/input-output-hk/cardano-ledger-specs/blob/master/shelley-ma/shelley-ma-test/cddl-files/shelley-ma.cddl) binary spec so we should ideally specify a nice way for UTXOs that contains enough info for it to be useful in all use-cases. What you have in this PR seems like a solid choice but we should update the spec first as it currently references the following cddl:
```cddl
transaction_output = [address, amount : value]
```

rust/src/utils.rs
```diff
@@ -64,6 +64,86 @@ macro_rules! to_from_bytes {
     }
 }
 
+#[wasm_bindgen]
+#[derive(Clone, Debug)]
+pub struct TransactionUnspentOutput {
+    input: TransactionInput,
+    value: Value,
```
SebastienGllmt, line 71: Is there a reason to specify `value` and `address` instead of just using `output: TransactionOutput`?

alessandrokonrad: I oversaw that with **`TransactionOutput`**, your solution makes more sense right.

rooooooooob: I updated the CIP with this change using the cddl
```
transaction_unspent_output = [
        input: transaction_input,
        output: transaction_output,
]
```

PR #169, opened 2021-06-16
run rustfmt

islishude:
rustfmt rust/**/*.rs


SebastienGllmt:
Parts of this repository are created with codegen tools that don't conform to rustfmt which is why we haven't done this before. Running rustfmt would just cause diffs whenever we re-regenerate these files.

PR #175, opened 2021-07-08
Alonzo: BigInt wrapper type

rooooooooob:
CDDL defines:
```
int = uint / nint
biguint = #6.2(bstr)
bignint = #6.3(bstr)
bigint = biguint / bignint
integer = int / bigint
unsigned = uint / biguint
```

which is very unwieldly to use and not to mention to have to interpret
the bytes correctly and everything so this commit introduces a single
wrapper type `BigInt` that covers all of these cases and tries to encode
the integer in the smallest possible type.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -16,10 +16,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1d49d90015b3c36167a20fe2810c5cd875ad504b39cff3d4eae7977e6b7c1cb2"
 
 [[package]]
-name = "bech32"
-version = "0.7.3"
+name = "autocfg"
+version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2dabbe35f96fb9507f7330793dc490461b2962659ac5d427181e451a623751d1"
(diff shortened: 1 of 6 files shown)


rooooooooob:
Note: one area where this could cause problems if anyone needs canonical CBOR and wants to represent a number between -2^63 - 1 and -2^64 (whatever doesn't fit in i64 but would fit in a 64-byte entirely-negaitve number) we run into the usual restriction with the cbor_event CBOR library which I guess for ease of use forces users to go through an `i64` despite that not covering the entirety of the CDDL `nint` type...

That's why one of the CBOR RFC example test cases is commented out.

IIRC we had a user complaining about this in metadata about not handling integers in this specific range in some issue someone brought up but I can't find it.

rooooooooob:
Also I did not add in arithmetic types onto this BigInt since I don't think users are going to need to use them. If they really want to do arithemtic they can use `to_str()` then parse it into JS's BigInt or whichever BigInt lib they want.

This class is uses by:

- Plutus integers (afaik this is compiled plutus core serialized structures so I don't think people will be messing with this much?)
- Cost models: maybe if someone wants to calculate this on top of our library? But that would likely involve another library to interpret the plutus for computing costs

If anyone thinks we should add them I can expose them though

stackchain (approved):
LGTM

vsubhuman (approved):

PR #179, opened 2021-07-22
Fix Shelley specs link

katomm:
Replace docs.cardano.org with link to cardano-ledger-specs since the content seems to no longer exists on docs.


stackchain (approved):

vsubhuman:
Thank you, @katomm!

PR #183, opened 2021-07-23
remove non necessary dependencies from the rust crate

NicolasDP:
I fail to see where this is needed in the code. This is cleaning up 9 dependencies that were not used (or so it seems ?).


vsubhuman:
Thank you, @NicolasDP!

rooooooooob (commented):
We used to use entire crates from chain-libs from jormungandr (chain-crypto, chain-legacy-addresses, etc) and those dependencies were added when we instead just added the parts we needed in PR #61 to reduce the binary size I think? I would assume they were at least at the time needed to power some parts of the `chain-crypto` code we were using beyond `cryptoxide`, but maybe they were either unnecessarily added or we removed code using those since then. If this still compiles in rust / wasm builds it should be fine to remove them I guess.

stackchain (approved):

PR #187, opened 2021-08-11
TxBuilder: max output size + multiple change address support

rooooooooob:
Users with many NFTs could run into problems creating certain txs that
returned many of them as a change output as it would surpass the maximum
output size limit in the ledger. The TxBuilder now takes this limit into
account and errors if it is passed, and during the change output
creation will attempt to spread out the assets over multiple outputs to
not surpass the ledger limit.

This is done as a greedy selection for now as the limit is currently 4kb
or so and each asset is at most 64 bytes each so this should suffice.

### DIFF
--- a/rust/src/tx_builder.rs
+++ b/rust/src/tx_builder.rs
@@ -121,6 +121,7 @@ pub struct TransactionBuilder {
     minimum_utxo_val: BigNum,
     pool_deposit: BigNum,
     key_deposit: BigNum,
+    max_output_size: u32,
     fee_algo: fees::LinearFee,
     inputs: Vec<TxBuilderInput>,
     outputs: TransactionOutputs,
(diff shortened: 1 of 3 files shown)


rust/src/tx_builder.rs
```diff
@@ -473,7 +544,19 @@ impl TransactionBuilder {
...
+        let full_tx = fake_full_tx(self, built)?;
+        let full_tx_size = full_tx.to_bytes().len();
+        if full_tx_size > self.max_tx_size as usize {
+            Err(JsError::from_str(&format!(
+                "Maximum transaction size of {} exceeded. Found: {}",
+                self.max_tx_size,
+                full_tx_size
+            )))
+        } else {
+            Ok(full_tx.body)
```
vsubhuman, line 558: This is similar to returning `Ok(built)`, right? Just looks weird that the result used to be that constructed tx body and now as if it is important for it to be taken from the fake `full_tx`

PR #190, opened 2021-08-17
Exclude empty multiasset struct when subtracting values

### LINKED ISSUE #189: Non-token outputs seem to sometimes contain empty token bundle
We got a report in ledgerjs repo that a user of cardano-serialization-lib with a transaction containing tokens only in some of its outputs had an empty token bundle in all of them when serialized, which then breaks compatibility with hw wallets which never add this empty token bundle to non-token outputs. Is including an empty token bundle in the CBOR-serialized outputs intended?

Link to related issue: https://github.com/vacuumlabs/ledgerjs-cardano-shelley/issues/125

SebastienGllmt:
This fixes #189

### DIFF
--- a/rust/src/tx_builder.rs
+++ b/rust/src/tx_builder.rs
@@ -1188,4 +1188,65 @@ mod tests {
         );
         let _final_tx = tx_builder.build(); // just test that it doesn't throw
     }
+
+    #[test]
+    fn build_tx_no_useless_multiasset() {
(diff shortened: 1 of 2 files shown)


vsubhuman (approved):

vsubhuman (commented):
/check

vsubhuman (commented):
/check

PR #201, opened 2021-09-14
Hashes

alessandrokonrad:
I added the hash functions for datum and script data.

### DIFF
--- a/rust/src/plutus.rs
+++ b/rust/src/plutus.rs
@@ -375,6 +375,8 @@ impl PlutusData {
 #[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd)]
 pub struct PlutusList(Vec<PlutusData>);
 
+to_from_bytes!(PlutusList);
+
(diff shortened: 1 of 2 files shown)


alessandrokonrad:
I forgot to wrap the cost model inside language views:
https://github.com/input-output-hk/cardano-ledger-specs/blob/e59e6ae405e9054966180a26d759b078f8f87329/alonzo/test/cddl-files/alonzo.cddl#L97

Will update that

vsubhuman:
Thank you, @alessandrokonrad! Reviewing it now

alessandrokonrad:
> Thank you, @alessandrokonrad! Reviewing it now

Awesome. I'm still struggeling a little bit with the hash_script_data function. Was yet not able to reproduce the same hash in in the serialization-lib as in the cardano-cli, although I feel according to the specs it looks correct. The concatenation of redeemers, datums and cost models.

vsubhuman (approved):

PR #203, opened 2021-09-18
fixed serializer for CostModel

alessandrokonrad:
CostModel is no longer a map and it was probably just forgotten to change from map to array.


vsubhuman (approved):
Thank you, @alessandrokonrad!

Issue #16, opened 2020-07-08
the transaction created by libs can't send to cardano-node


islishude:
shelley cardano-node: v1.14.2
shelley cardano-wallet: 2020.7.6-shelley

code with commit 320191cc915b80c3016d4bfb62599491b61d5a8d

```rust
#![allow(unused_variables, unused_imports)]

use cardano_serialization_lib::{self as adalib, crypto};
use cbor_event::{self as cbor};

pub fn main() {
    let xprv = "xprv1wq9th6qy7ej8w03cgas2lzwvzncvwaje8ek8lps6u6mwyhtpaewap0w2f9hhrpg6tvrffcvz5ecevvhfl6vnznswss96kqyncm2drut3kfps0pv0r4q3fuy6urzs52ywdpzx55muncgf283fr378hwzwp5p27ewe";
    let xprv = crypto::Bip32PrivateKey::from_bech32(xprv).unwrap();

    let xpub = xprv.to_public();
    // crypto::blake2b224 visibility changed
    let keyhash = crypto::blake2b224(&xpub.as_bytes()[0..32]).to_vec();
    let keyhash = crypto::AddrKeyHash::from_bytes(keyhash).unwrap();
    let cred = adalib::address::StakeCredential::from_keyhash(&keyhash);
    let payment = adalib::address::EnterpriseAddress::new(0, &cred);

    let mut inputs = adalib::TransactionInputs::new();
    let txid = "2558aad25ec6b0e74009f36dc60d7fec6602ce43d603e80c9edde9dd54c78eb4";
    let txid = crypto::TransactionHash::from_bytes(hex::decode(txid).unwrap()).unwrap();
    let input = adalib::TransactionInput::new(&txid, 0);
    inputs.add(&input);

    let mut outputs = adalib::TransactionOutputs::new();
    let output = adalib::TransactionOutput::new(&payment.to_address(), adalib::Coin::new(9000000));
    outputs.add(&output);

    let txbody =
        adalib::TransactionBody::new(&inputs, &outputs, adalib::Coin::new(1000000), 832163);

    let vkwt = txbody.sign(&xprv.to_raw_key());
    let mut vkwts = crypto::Vkeywitnesses::new();
    vkwts.add(&vkwt);
    let mut witness_set = adalib::TransactionWitnessSet::new();
    witness_set.set_vkeys(&vkwts);
    let newtx = adalib::Transaction::new(&txbody, &witness_set, None);

[...]

PR #211, opened 2021-09-24
Fix asset size calculation (#16)

### LINKED ISSUE #194: Minimum UTxO calculation missmatch (7.1.0)
It has been reported by users that for certain utxo's, often containing a large number of tokens, that the way cardano-serialization-lib calculates the minimum amount of Lovelace that output utxo is required to contain, doesn't match with our calculation.

I haven't found a way to replicate it using one of my own wallets but I was given access to a wallet in read-only(pub key) mode where I was able to replicate the issue. It's not a single occurrence either but has been reported by multiple users with large NFT collections. It's not about the number of tokens included in the output though as I have been able to construct large outputs without issue where the calculation from csl matches our calculation down to the Lovelace. Checked by first outputting the value from our calculation and then trying to construct the output with one Lovelace and watching the response from the error message where it prints the required amount. My guess is that it's related to the format of the asset name somehow. 

Our method to calculate minimum utxo has also been used in CLI tools for a long time without a single user reporting issues with it. 

The calculation is based on this documentation:
https://github.com/input-output-hk/cardano-ledger-specs/blob/master/doc/explanations/min-utxo.rst

And here is our code. It might be an error in our code but in that case, I cant spot it. 

```javascript
export const getMinUtxo       = (protoParams: IProtocolParams, tokens: IToken[]): string => {

  // using Number data type internally (fine for numbers of this size)

  let minUtxo                 = protoParams.minUTxOValue

  if(tokens.length === 0) return minUtxo.toString()

  // chain constants, based on: https://hydra.iohk.io/build/5949624/download/1/shelley-ma.pdf
  const k0                    = 0   // TODO: coinSize=0 in mary-era, 2 in alonzo-era
  const k1                    = 6
  const k2                    = 12  // assetSize=12
  const k3                    = 28  // pidSize=28
  const k4                    = 8   // word=8 bytes
  const utxoEntrySizeWithoutVal = 27  // 6 + txOutLenNoVal(14) + txInLen(7)
  const adaOnlyUTxOSize = utxoEntrySizeWithoutVal + k0

  const pidCollector          = []
  const assetsCollector       = []
  const nameCollector         = []

  for(const token of tokens) {

    if(pidCollector.indexOf(token.policy) === -1) pidCollector.push(token.policy)
    if(assetsCollector.indexOf(token.policy + token.name) === -1) assetsCollector.push(token.policy + token.name)
    if(token.name.length > 0 && nameCollector.indexOf(token.name) === -1) nameCollector.push(token.name)
  }

  const sumAssetNameLengths   = nameCollector.reduce((acc, val) => ( acc + (val.length/2) ), 0)
  const roundupBytesToWords   = Math.floor(((assetsCollector.length * k2) + sumAssetNameLengths + (pidCollector.length*k3) + (k4-1)) / k4)
  const tokenBundleSize       = k1 + roundupBytesToWords
  const minAda                = Math.floor(protoParams.minUTxOValue / adaOnlyUTxOSize) * (utxoEntrySizeWithoutVal + tokenBundleSize)

  if(minAda > Number(minUtxo)) { minUtxo = minAda }

  return minUtxo.toString()
}
```

rooooooooob:
* fix: asset size calculation [ch592]

* test: start a test using the inputs from the example [ch592]

* test: add test Value from #194 [ch592]

Fixes #194


vsubhuman (commented):
/check

vsubhuman (approved):

PR #214, opened 2021-09-29
fake_full_tx in TransactionBuilder should use static data #213

CantTouchDis:
To improve the performance of `fake_full_tx` I changed the call to `from_bip39_entropy` to `from_bytes` as mentioned by @rooooooooob and @SebastienGllmt in #213

Not much of a rust guy, but the tests passed. And my small code example ran infinitely faster than the previous attempts.

PR #215, opened 2021-09-29
Extracted fake-priv-key into a function and added tests

vsubhuman:
Preparing to close: https://github.com/Emurgo/cardano-serialization-lib/pull/214


vsubhuman (commented):
/check

Issue #227, opened 2021-10-22
PlutusList array serialization is different from cardano-cli


kodemill:
Currently this library uses the **fixed-length array** CBOR type to serialize the `PlutusList` data type, whereas from what we've observed the cardano-cli encodes the datum as [**indefinite-length array**](https://www.rfc-editor.org/rfc/rfc8949.html#name-indefinite-length-arrays-an) with some exceptions:

- an empty array is encoded as fixed (zero) length array
- the root level PlutusList in the TransactionWitnessSet is encoded as a fixed length array

An example hex-encoded cbor of a datum generated by the cli:
`d8799f4100d8799fd8799fd8799f581cffffffffffffffffffffffffffffffffffffffffffffffffffffffffffd8799fd8799fd8799f581cffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffd87a80ff1a002625a0d8799fd879801a000f4240d87a80ffff`
Decoding and then re-encoding the above datum using `PlutusData` from this lib will result in the following:
`d879844100d87982d87982d87981581cffffffffffffffffffffffffffffffffffffffffffffffffffffffffd87981d87981d87981581cffffffffffffffffffffffffffffffffffffffffffffffffffffffffd87a801a002625a0d87983d879801a000f4240d87a80`

This minor inconsistency between these core tools might cause some problems for the ecosystem developers. I'd be happy to provide a PR to fix this.

As an example, when a wallet signs a transaction and hands it back, the hash of the datum is different, leading to `NonOutputSupplimentaryDatums` errors when submitting the transaction.

JaredCorduan:
thank you @kodemill . I'll make a link here to the ledger issue: https://github.com/input-output-hk/cardano-ledger-specs/issues/2512

rooooooooob:
Closing due to #228 being merged.

mailtodanish:
We generate datum script using below :

cardano-cli transaction hash-script-data --script-data-file ${datum_name}.datum

It generates the datum script. once we have datum we need to incluide it into transaction 
that is generated on cli again


 cardano-cli transaction build \
                --alonzo-era \
                --tx-in ${utxoin} \
                --tx-out ${SCRIPT_ADDR}+${LOCK_AMNT} \
                --tx-out-datum-hash ${hashes[0]} \
                --metadata-json-file metadata.json \
				
Please let me know how to do above datum using nami-wallet-api or

cardano-serialization-lib-browser ? any reference

[...]

PR #217, opened 2021-10-06
Fixed hash script function (updated)

rooooooooob:
continues with #209 to fix `utils::hash_script_data()` which in the Haskell node instead of using `cost_mdls` directly seems to construct some kind of canonical `LangViewDep` structure to hash instead.

PR #228, opened 2021-10-22
Fix plutus list serialization

kodemill:
#227
- serialize `PlutusList` as indefinite length array instead of fixed length array, with the exception of 0-length arrays
- serialize the `PlutusData` (PlutusList) in the TransactionWitnessSet as fixed length arrays


JaredCorduan:
Our hope in the ledger is to fix the serialization bug in the next hard fork https://github.com/input-output-hk/cardano-ledger-specs/issues/2512

BlakeBrown:
Just a casual observer here, how do we know if our datum hash is correct if it differs between the cli and the serialization lib? Is there no way to test this until the issue gets resolved in the next hard fork?

rooooooooob:
>Just a casual observer here, how do we know if our datum hash is correct if it differs between the cli and the serialization lib? Is there no way to test this until the issue gets resolved in the next hard fork?

Unfortunately that's why hashing is so useful - small changes in the input cause huge changes in the output so you can't predict it. This makes debugging not very nice about why the hash is wrong. You just know that **something** is different/wrong. If you can export the data that the cardano-node hashes that can help a lot since we can manually examine that and compare it with the cardano-serialization-lib's structure. Any difference is probably something small like the CBOR encoding differences (e.g. definite vs indefinite list length encoding) here or in #217 for example. Or even just having any similar sample data from the node vs ours can help since you can look out for small differences in the CBOR encoding like that. That's how I figured it out in #217 by looking at sample data from the cardano-node and noticing the indefinite CBOR list/map length encoding instead of definite.

vsubhuman:
> Our hope in the ledger is to fix the serialization bug in the next hard fork [input-output-hk/cardano-ledger#2512](https://github.com/input-output-hk/cardano-ledger/issues/2512)

Am I getting it right that if we merge this change right now for the hardfork we will have to change it back?

JaredCorduan:
> > Our hope in the ledger is to fix the serialization bug in the next hard fork [input-output-hk/cardano-ledger#2512](https://github.com/input-output-hk/cardano-ledger/issues/2512)
> 
> Am I getting it right that if we merge this change right now for the hardfork we will have to change it back?

Plutus V1 is domed to have to always use this sad script integrity hash, but our hope is to have Plutus V2 use what we had intended.

vsubhuman:
@kodemill , can you plz pull latest master and also run the `npm run rust:test` and resolve any test failures?

I am getting this failure when checking your branch locally, and it does not happen on master:

```

failures:

---- utils::tests::correct_script_data_hash stdout ----
language_views = a141005901d59f1a000302590001011a00060bc719026d00011a000249f01903e800011a000249f018201a0025cea81971f70419744d186419744d186419744d186419744d186419744d186419744d18641864186419744d18641a000249f018201a000249f018201a000249f018201a000249f01903e800011a000249f018201a000249f01903e800081a000242201a00067e2318760001011a000249f01903e800081a000249f01a0001b79818f7011a000249f0192710011a0002155e19052e011903e81a000249f01903e8011a000249f018201a000249f018201a000249f0182001011a000249f0011a000249f0041a000194af18f8011a000194af18f8011a0002377c190556011a0002bdea1901f1011a000249f018201a000249f018201a000249f018201a000249f018201a000249f018201a000249f018201a000242201a00067e23187600010119f04c192bd200011a000249f018201a000242201a00067e2318760001011a000242201a00067e2318760001011a0025cea81971f704001a000141bb041a000249f019138800011a000249f018201a000302590001011a000249f018201a000249f018201a000249f018201a000249f018201a000249f018201a000249f018201a000249f018201a00330da70101ff
thread 'utils::tests::correct_script_data_hash' panicked at 'assertion failed: `(left == right)`
  left: `"4415e6667e6d6bbd992af5092d48e3c2ba9825200d0234d2470068f7f0f178b3"`,
 right: `"57240d358f8ab6128c4a66340271e4fec39b4971232add308f01a5809313adcf"`', src/utils.rs:2100:9


failures:
    utils::tests::correct_script_data_hash

test result: FAILED. 109 passed; 1 failed; 1 ignored; 0 measured; 0 filtered out

```

kodemill:
@vsubhuman I've fixed the failing test and added a couple more to cover the introduced changes in this PR.

vsubhuman:
> I've fixed the failing test and added a couple more to cover the changes introduced in this PR.

Thank you, @kodemill !

vsubhuman (approved):

PR #232, opened 2021-10-27
TxBuilder: Implement CIP-0002 Coin Selection

rooooooooob:
https://github.com/cardano-foundation/CIPs/blob/master/CIP-0002/CIP-0002.md

key differences
1) we take into account adjuting for free (which is out of CIP2's scope)
2) LargestFirst supports multiassets (naively)


pedromtcosta (approved):
Looks good! Nice tests

vsubhuman (commented):
/check

vsubhuman (approved):

rooooooooob:
@BlakeBrown please see #264 which implements multiasset for both. The current (i.e. this PR) implementation for multiasset is very sub-optimal and will almost certainly add too many inputs in unless you have a lot of inputs with those desired multiassets. I should have thrown an error for the CIP-2 random-improve if there were multiassets too, and I do now in #264 if you're using the non-multiasset enum variants.

Keep in mind CIP-2 only targets ADA-only since it was written before Cardano had multiassets. I tried to adapt both to multiasset in the linked PR.

PR #236, opened 2021-10-29
Split policy ID assets when overflow

pedromtcosta:
# Abstract
This PR makes changes the `add_change_if_needed` function from the `TransactionBuilder` class to split assets among multiple outputs when generating a change output which contains a large number of assets.

# Motivation
While unlikely, it is possible that we run into an overflow on the size of the output when generating a change outputs which contains several assets. This can be handled by splitting these assets among multiple outputs.

# Background
When we create a new instance of `TransactionBuilder` we specify a value for the `max_value_size`, which is a size limit for the value of each output in the transaction. If while building the change output it gets larger than `max_value_size`, the transaction building will result in an error.
See the [task description on Asana](https://app.asana.com/0/1200842454080452/1201160407953375).

# Implementation
To handle the scenario described above, instead of just inserting all the assets into the change output directly, we loop through all the asset names from the `Assets` object which comes from the change estimator `MultiAsset` and insert them one by one into a new `Assets` object, defined as `rebuilt_assets`, and only afterwards we insert this `rebuilt_assets` object into the `MultiAsset` and then finally we add the `MultiAsset` into the output.

Before inserting each individual asset into the `rebuilt_assets` object, we simulate an insert with cloned objects to verify if adding this asset to the output will cause an overflow, by using the newly defined `will_adding_asset_make_output_overflow` function. If it doesn't cause the overflow, we just continue normally. If it ***does*** cause an overflow, we insert the current `rebuilt_assets` into the output and generate a new one and continue inserting the remaining assets into this new output.

We store all the `MultiAsset` objects from the outputs generated by the `pack_nfts_for_change` function into a `Vec<MultiAsset>` and change `pack_nfts_for_change` to return this vector instead of a single `MultiAsset`. Finally, we use the same logic for adding the NFT change we had before for each of the items returned by `pack_nfts_for_change`


vsubhuman:
@pedromtcosta , please resolve conflicts

pedromtcosta:
@vsubhuman Conflicts resolved

vsubhuman:
@ozgrakkurt plz review

vsubhuman:
@rooooooooob , check plz if you have a chance

rust/src/tx_builder.rs
```diff
@@ -814,9 +814,25 @@ impl TransactionBuilder {
                 }
                 let change_estimator = input_total.checked_sub(&output_total)?;
                 if has_assets(change_estimator.multiasset()) {
-                    fn pack_nfts_for_change(max_value_size: u32, change_address: &Address, change_estimator: &Value) -> Result<MultiAsset, JsError> {
+                    fn will_adding_asset_make_output_overflow(output: &TransactionOutput, current_assets: &Assets, asset_to_add: (PolicyID, AssetName, BigNum), max_value_size: u32) -> bool {
+                        let (policy, asset_name, value) = asset_to_add;
+                        let mut current_assets_clone = current_assets.clone();
+                        current_assets_clone.insert(&asset_name, &value);
+                        let mut amount_clone = output.amount.clone();
```
rooooooooob, line 821: We should be able to avoid this clone but it's nothing important

rooooooooob (approved):
LGTM

rooooooooob:
We should squash and merge this instead of a single merge commit IMO. We've been having so many merging master into PR branches lately and the commit history has been getting really dirty. It might be worth doing that in general now.

PR #242, opened 2021-11-02
Add StakeCredential enum type

SebastienGllmt:
Previously there was no way to get proper type-checking working on the enum from JS since the function just returned `number`. Now it properly returns an enum so you can type-check on it (this is the same thing we do with CertificateKind, etc.)

### DIFF
--- a/rust/src/address.rs
+++ b/rust/src/address.rs
@@ -71,6 +71,14 @@ enum StakeCredType {
 }
 
 #[wasm_bindgen]
+#[repr(u8)]
+#[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd)]
+pub enum StakeCredKind {
+    Key,
+    Script,
+}
+
+#[wasm_bindgen]
 #[derive(Debug, Clone, Eq, Hash, Ord, PartialEq, PartialOrd)]
 pub struct StakeCredential(StakeCredType);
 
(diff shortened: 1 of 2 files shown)


rust/src/address.rs
```diff
@@ -98,10 +106,10 @@ impl StakeCredential {
         }
     }
 
-    pub fn kind(&self) -> u8 {
+    pub fn kind(&self) -> StakeCredKind {
         match &self.0 {
-            StakeCredType::Key(_) => 0,
-            StakeCredType::Script(_) => 1,
+            StakeCredType::Key(_) => StakeCredKind::Key,
+            StakeCredType::Script(_) => StakeCredKind::Script,
```
vsubhuman, line 112: Thanks, Seba! Just to double-check In JS it will be returning the **values** from `declare export var StakeCredKind` so backward compatible, right?

SebastienGllmt: yes, backwards compatible for JS which is why the return type of the Flow definition didn't change. Not backwards compatible for Rust, but that shouldn't be an issue

vsubhuman (approved):

rust/src/address.rs
vsubhuman: Perfect

vsubhuman (commented):
/check

Issue #240, opened 2021-11-01
ledgerjs-hw-app-cardano 4.0.0 contains a breaking change for CSL


Scitz0:
ledgerjs-hw-app-cardano 4.0.0 was officially released a week ago. This brings a couple of key changes, most importantly:

> validation of canonical ordering of cbor map keys compliant with [CIP 21](https://cips.cardano.org/cips/cip21) (withdrawals, token policy ids in outputs and mint, asset names within an asset group)

The validation of canonical ordering in ledgerjs first sort on asset name length, and secondly if equal length, string comparision.
Snippet of code validation done:
```
const sortedAssetNames = [...assetNamesHex].sort( (n1, n2) => {
  if (n1.length === n2.length) return n1.localeCompare(n2)
  else return n1.length - n2.length
})
validate(JSON.stringify(assetNamesHex) === JSON.stringify(sortedAssetNames), InvalidDataReason.MULTIASSET_INVALID_ASSET_GROUP_ORDERING)
```

This means that for three assets in the same policy called `AA`, `B` and `CC`, according to ledgerjs they should be sorted as `B > AA > CC`. But if you insert tokens in this order into a CSL Value object they are resorted as `AA > B > CC`. 

Due to the above, deployment of ledgerjs-hw-app-cardano 4.0.0 breaks compatibility with the current version of CSL. Ledgerjs 4.0.0 is mandatory for Ledger Cardano App 3.0.0 support already into the wild (albeit by mistake it seems, now pulled again). 

Can we get this issue prioretized?

vsubhuman:
Thank you, @Scitz0 !

vsubhuman:
Ok, @Scitz0, do you think then it would be enough for the `MultiAsset` to preserve the order of the asset names as they are being inserted, without forcing the new way of sorting?

MarcelKlammer:
I also think, that sorting can be done by the lib user before inserting.

Scitz0:
> Ok, @Scitz0, do you think then it would be enough for the `MultiAsset` to preserve the order of the asset names as they are being inserted, without forcing the new way of sorting?

Yip, this would be my preference.

gitmachtl:
This is also related to https://github.com/vacuumlabs/cardano-hw-cli/issues/104
[...]

PR #243, opened 2021-11-02
Switched `Assets` and `MultiAsset` to use `LinkedHashMap`

vsubhuman:
For compatibility with new Ledger requirements the client apps must be able to preserve a certain insertion order for asset policies and asset names.

See: https://github.com/Emurgo/cardano-serialization-lib/issues/240

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -58,6 +58,7 @@ use plutus::*;
 use metadata::*;
 use utils::*;
 use linked_hash_map::LinkedHashMap;
+use itertools::Itertools;
 
 type DeltaCoin = Int;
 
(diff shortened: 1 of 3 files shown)


rooooooooob (approved):
LGTM from an entirely code-related perspective. I'm still not sure if this is the approach we want. We're going to want to document this properly that the user needs to insert in the correct order and where that is defined.

Why is preserving the insert order of importance if the node only accepts the canonical ordering? I feel like this is going to cause users pain when they don't insert in the right order.

vsubhuman:
This change is rejected to rather implement the canonical sorting from [CIP21](https://cips.cardano.org/cips/cip21/) directly

Issue #237, opened 2021-10-30
ASM lib - PrivateKey.from_bech32 is undefined


34r7h:
```
import * as CW from "./node_modules/@emurgo/cardano-serialization-lib-asmjs/cardano_serialization_lib.js"
console.log(CW.PrivateKey.from_bech32)
```
undefined

I'm trying to add bech32 js library to convert but not sure how to implement the other methods like .toPublic()

rooooooooob:
this should be fixed by #244

Issue #238, opened 2021-10-30
Bip32PrivateKey.to_raw_key, Bip32PrivateKey.to_public - Error: null pointer passed to rust


34r7h:
Following up on #237, I think the docs are known, for quite some time, to be misleading. I followed an issue from February, #120 but still can't make sense of this problem. 

The class prototype has the methods. Is there a rust/wasm setup step somewhere? Perhaps another installation related soultion?

**package.json**
```
{
  "name": "serializer",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "@emurgo/cardano-serialization-lib-asmjs": "^9.1.0",
    "@emurgo/cardano-serialization-lib-nodejs": "^9.1.0",
    "bech32": "^2.0.0",
    "ed25519-hd-key": "^1.2.0"
  },
  "type": "module"
}
```
**index.js**
```
import * as CW from "./node_modules/@emurgo/cardano-serialization-lib-asmjs/cardano_serialization_lib.js"
import {bech32} from 'bech32'

const okey = 'ed25519e_sk16rl5fqqf4mg27syjzjrq8h3vq44jnnv52mvyzdttldszjj7a64xtmjwgjtfy25lu0xmv40306lj9pcqpa6slry9eh3mtlqvfjz93vuq0grl80';
const prvKey = new CW.Bip32PrivateKey(bech32.decode(okey, 200));
console.log(
    prvKey.to_public.toString(), 
    Object.getOwnPropertyNames(
        Object.getPrototypeOf(prvKey)
    )
);

```
OUTPUT
```
to_public() {
        var ret = wasm.bip32privatekey_to_public(this.ptr);
        return Bip32PublicKey.__wrap(ret);
    } [
  'constructor',
  '__destroy_into_raw',
  'free',
  'derive',
  'to_128_xprv',
  'to_raw_key',
  'to_public',
  'as_bytes',
  'to_bech32',
  'chaincode'
]

file:///home/34r7h/serializer/node_modules/@emurgo/cardano-serialization-lib-asmjs/cardano_serialization_lib_bg.js:12145
    throw new Error(getStringFromWasm0(arg0, arg1));
          ^

Error: null pointer passed to rust
[...]

Issue #120, opened 2021-02-13
How to use `Address` instead of `ByronAddress` as witness


faustbrian:
I am currently looking at https://github.com/Emurgo/cardano-serialization-lib/blob/master/example/index.spec.ts#L75 and https://github.com/Emurgo/cardano-serialization-lib/blob/master/example/index.spec.ts#L104 and trying to use a Shelley address for the same process by using `Address.from_bech32` but the witness methods all expect an instance of `ByronAddress`.

Is support for Shelley addresses fully implemented and if so, what would be the process of using a Shelley address as a witness?

SebastienGllmt:
Shelley addresses (usually `BaseAddress` witnesses are created using `make_vkey_witness` (https://github.com/Emurgo/cardano-serialization-lib/blob/15d193f097618e8dae5c03d5d711da9e08c4f706/rust/src/utils.rs#L530)

And yes, all of Shelley (and every other hardfork) is supported.

You can find example code here (but this example code hasn't been updated for Mary yet) https://github.com/Emurgo/cardano-serialization-lib/blob/59426b43d105e6cf4f05e4b5945c0cb3dc483130/doc/getting-started/generating-transactions.md#example-code

faustbrian:
Thanks, that got me a bit further but the `CardanoWasm.PrivateKey.from_bech32` method doesn't seem to exist anymore which is used in `const prvKey = CardanoWasm.PrivateKey.from_bech32("ed25519e_sk16rl5fqqf4mg27syjzjrq8h3vq44jnnv52mvyzdttldszjj7a64xtmjwgjtfy25lu0xmv40306lj9pcqpa6slry9eh3mtlqvfjz93vuq0grl80");`. What has this been replaced with?

CherryDT:
@faustbrian Why did you stike-through your text? What solution did you find? I still cannot find a method to convert the result of `privateKey.to_bech32()` back to a `PrivateKey`. The [generating transactions example](https://docs.cardano.org/projects/cardano-serialization-lib/en/latest/getting-started/generating-keys.html) also shows the use of `CardanoWasm.PrivateKey.from_bech32` which is undefined...

faustbrian:
@CherryDT our solution is here https://github.com/ArkEcosystem/platform-sdk/blob/master/packages/platform-sdk-ada/src/services/transaction.ts#L26-L136

CherryDT:
[...]

PR #244, opened 2021-11-03
Added PrivateKey::from_bech32()

rooooooooob:
This method existed already in js-chain-libs where we got this class
from to begin wtih. It also exists in the other 3 keys types:
PublicKey, Bip32PublicKey and Bip32PrivateKey so it makes sense to add here.

[taken from here](https://github.com/input-output-hk/js-chain-libs/blob/master/src/lib.rs#L47)

I'm not sure why it wasn't there to begin with. I didn't see it being removed in the git histories either yet it was in js-chain-libs.

### DIFF
--- a/rust/pkg/cardano_serialization_lib.js.flow
+++ b/rust/pkg/cardano_serialization_lib.js.flow
@@ -3382,6 +3382,12 @@ declare export class PrivateKey {
   static generate_ed25519extended(): PrivateKey;
 
   /**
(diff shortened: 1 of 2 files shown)


rooooooooob:
This was mentioned in issue #238 and also #120

rooooooooob:
It should fix #237 directly

vsubhuman (approved):
Thanks, Rob!

PR #245, opened 2021-11-03
Added a test for new `PrivateKey::from_bech32` function

vsubhuman:
Noticed too late that https://github.com/Emurgo/cardano-serialization-lib/pull/244 is missing a test


vsubhuman:
@rooooooooob , check when possible

rooooooooob (approved):
LGTM. We could add in the one from the doc string comment in the function but it should be fine. I got the code directly from `js-chain-libs` which is where `PrivateKey` came from originally anyway.

vsubhuman (commented):
Perfect! /check

PR #246, opened 2021-11-03
[9.1.1] Implemented canonical ordering for `AssetName` and added a test

vsubhuman:
As described in CIP21: https://cips.cardano.org/cips/cip21/

> **Multiassets**
>
> Since multiassets (policy_id and asset_name) are represented as maps, both need to be sorted in accordance with the specified canonical CBOR format. Also, an output or the mint field must not contain duplicate policy_ids and a policy must not contain duplicate asset_names.

Policy IDs are always same length, so default lexicographical ordering is correct, but asset names might be different length, so they require the sorting to be implemented explicitly.

Canonical CBOR is defined in section 3.9 here: https://www.rfc-editor.org/rfc/rfc7049.txt

> The keys in every map must be sorted lowest value to highest.
      Sorting is performed on the bytes of the representation of the key
      data items without paying attention to the 3/5 bit splitting for
      major types.  (Note that this rule allows maps that have keys of
      different types, even though that is probably a bad practice that
      could lead to errors in some canonicalization implementations.)
      The sorting rules are:
>
>      *  If two keys have different lengths,
>          the shorter one sorts earlier;
>
>      *  If two keys have the same length,
>          the one with the lower value in (byte-wise) lexical order sorts earlier.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "cardano-serialization-lib",
-  "version": "9.1.0",
+  "version": "9.1.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cardano-serialization-lib",
-  "version": "9.1.0",
+  "version": "9.1.1",
   "description": "(De)serialization functions for the Cardano blockchain along with related utility functions",
   "scripts": {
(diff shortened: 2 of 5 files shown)


rooooooooob (approved):
LGTM

vsubhuman:
Release: https://github.com/Emurgo/cardano-serialization-lib/releases/tag/9.1.1

New version published:
- https://www.npmjs.com/package/@emurgo/cardano-serialization-lib-browser/v/9.1.1
- https://www.npmjs.com/package/@emurgo/cardano-serialization-lib-nodejs/v/9.1.1
- https://www.npmjs.com/package/@emurgo/cardano-serialization-lib-asmjs/v/9.1.1

PR #250, opened 2021-11-06
PlutusConstrData rework tag parameter to be easier

### LINKED ISSUE #247: Inconsistent Data Hashing from CLI & Serialization Lib?
Hey there!

having a similar issue to some other issues open, but i'm serializing an **empty** datum with just the constructor set but yet the hashes still don't line up. See examples below, almost no data is being passed through besides the `tag`. 

to be specific:

serialization lib:
```javascript
  const datum = Loader.Cardano.PlutusData.new_constr_plutus_data(
    Loader.Cardano.ConstrPlutusData.new(
      Loader.Cardano.Int.new_i32(0),
      Loader.Cardano.PlutusList.new()
    )
  );
```

then on the cli
```bash
cardano-cli transaction hash-script-data \
  --script-data-file scripts/datum.json 
```

where datum.json is
```json
{"constructor":0,"fields":[]}
```

i may be overlooking something but do let me know! otherwise i can't seem to figure out where the difference is.

the hashes respectively are:
`f7f2f57c58b5e4872201ab678928b0d63935e82d022d385e1bad5bfe347e89d8`
`923918e403bf43c34b4ef6b48eb2ee04babed17320d8d1b9ff9ad086e86f44ec`

--- UPDATE: ---
for additional context i get the first hash by loading it into a `TransactionOutput` then calling `.data_hash().to_bytes()` then, hex encoding the byte output. the same output is achieved by just calling `hash_plutus_data()` then converting `.to_bytes()` then hex encoding.

rooooooooob:
Due to https://github.com/input-output-hk/plutus/blob/1f31e640e8a258185db01fa899da63f9018c0e85/plutus-core/plutus-core/src/PlutusCore/Data.hs#L61
we can see that unlike the direct CBOR representation, there is actually
a transformation done on the constructor alternative tag before
serialization which led to users being confused as the results differed
from what was being done by cardano-node.

We now expose only the conceptual constructor tag which is what the
cardano-node uses, rather than the wire-format specified in the CDDL
which is more of implementation detail.

Fixes #247

### DIFF
--- a/rust/src/plutus.rs
+++ b/rust/src/plutus.rs
@@ -52,7 +52,7 @@ impl PlutusScripts {
 #[wasm_bindgen]
 #[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd)]
 pub struct ConstrPlutusData {
-    tag: Int,
+    alternative: BigNum,
     data: PlutusList,
 }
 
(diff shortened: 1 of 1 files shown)


stephenhuh:
closes #247 - will close issue out! thanks again.

rust/src/plutus.rs
```diff
@@ -52,36 +52,60 @@ impl PlutusScripts {
 #[wasm_bindgen]
 #[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd)]
 pub struct ConstrPlutusData {
-    tag: Int,
+    alternative: BigNum,
```
stephenhuh, line 55: from what i can tell this will be a breaking change for any dapps currently using `tag` then

vsubhuman (approved):

PR #253, opened 2021-11-08
RandomImprove : panicked at 'cannot sample empty range'

vsubhuman:
Fixed an issue when random-improve uses all available inputs in the first loop and added a test-case for it

Before that it was failing with:
```
panicked at 'cannot sample empty range', <::core::macros::panic macros>:3:10
```

### DIFF
--- a/rust/src/tx_builder.rs
+++ b/rust/src/tx_builder.rs
@@ -220,21 +220,23 @@ impl TransactionBuilder {
                         associated_inputs.entry(output.clone()).or_default().push(input);
                     }
                 }
-                // Phase 2: Improvement
-                for output in outputs.iter_mut() {
-                    let associated = associated_inputs.get_mut(output).unwrap();
-                    for input in associated.iter_mut() {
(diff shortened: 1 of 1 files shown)


vsubhuman:
@rooooooooob , check plz when possible

vsubhuman:
@rooooooooob , here I have created a test case that fails consistently because there are three similar small available inputs, where two of them are enough to cover the output, but not enough to cover the output and the fee increased by adding these two inputs. Three of them would be enough to cover the output and the fees. But only two get added.

https://github.com/Emurgo/cardano-serialization-lib/pull/254

vsubhuman (commented):
/check

pedromtcosta (approved):

rooooooooob (approved):
LGTM, that was an oversight on my part since the other test didn't catch it.

vsubhuman:
> LGTM, that was an oversight on my part since the other test didn't catch it.

Yeah, I've also completely missed it on the review

Issue #252, opened 2021-11-08
Issue with the CIP2 random-improve coin-selection when small inputs and need to calculate change


vsubhuman:
The CIP2 coin-selection been added in this PR: https://github.com/Emurgo/cardano-serialization-lib/pull/232

See this fix for the random-improve test case where all amounts been multiplied by ten: https://github.com/Emurgo/cardano-serialization-lib/pull/251/commits/28f0988f3f387671a19aa07bc04a6b940bb60c1d#diff-67896f615f1e60f7474eb6194c0614a005522dc35fb07046855d77260954b972R2080-R2092

Without this fix, after there was a change to the minimum required coin calculation which made it impossible to just set the required number to `1` for test purposes, the test started failing **incosistently** with the `"Insufficient input in transaction"` error. And the reason for that is that random-improve branch of the coin selection works on the per-output basis and does not address the required fees explicitly.

You can see here that the fee for any new added input is being added to the `total_output` - https://github.com/Emurgo/cardano-serialization-lib/blob/28f0988f3f387671a19aa07bc04a6b940bb60c1d/rust/src/tx_builder.rs#L217

But then that `total_output` value is not used anywhere, there are only two loops going over outputs. So test was failing inconsistently because sometimes the random selection would get enough input to cover the fees, but sometimes, it doesn't.

vsubhuman:
The issue can be reproduced by changing the amounts in that test case back to what it was before, or adjusting them in a way that it will always fail.

vsubhuman:
@rooooooooob , here I have created a test case that fails consistently because there are three similar small available inputs, where two of them are enough to cover the output, but not enough to cover the output and the fee increased by adding these two inputs. Three of them would be enough to cover the output and the fees. But only two get added.

https://github.com/Emurgo/cardano-serialization-lib/pull/254

vsubhuman:
@pedromtcosta , can you plz look into this? Maybe we can do the fix if you will have some time available

pedromtcosta:
> @pedromtcosta , can you plz look into this? Maybe we can do the fix if you will have some time available
> 
> You can work right from this branch: #254 , to have the test right there

@vsubhuman I pushed some changes to this PR. Could you review them please?

PR #254, opened 2021-11-08
RandomImprove : Test covering fees correctly

vsubhuman:
Added test that consistently fails on random improve not adding enough funds to cover the fees


vsubhuman (commented):
/check

vsubhuman:
Fails with insufficient funds at the moment:

image

Waiting for https://github.com/Emurgo/cardano-serialization-lib/issues/252 to get resolved to fix that.

The conditions in the test are:
1. Output value is 100
2. Min fee when only output is added is 53
3. Three outputs are available, 150 each
4. When only two outputs are added (which is enough to cover the output) - the min fee is 228
5. The total then comes to 328 and two outputs are not enough to cover that - **hence the fail**
6. If three outputs would be added - min fee then is 264 and the total comes to 364
7. Three outputs together are 450 and should be enough to cover the required 364

pedromtcosta:
@vsubhuman I made the following changes:
- updated test to assert for the fee amount of `264` because now we expect the TX from that test to have 3 inputs;
- after "Phase 2: Improvement", if the total input is smaller than the total output, meaning the added fees from the inputs were not covered, we add the sufficient inputs to cover for the total output

rust/src/tx_builder.rs
```diff
@@ -240,6 +256,23 @@ impl TransactionBuilder {
                         }
                     }
                 }
+                // Phase 3: add extra inputs needed for fees
+                // We do this at the end because this new inputs won't be associated with
+                // a specific output, so the improvement algorithm we do above does not apply here.
+                if input_total.partial_cmp(&output_total).unwrap().eq(&Ordering::Less) {
```
vsubhuman, line 262: 1. Why using `partial_cmp` here?
2. Unchecked unwrap doesn't look good.

If these two types are both `Value` I think you can even just use `<`, e.g. see how `while added < needed` is used above

pedromtcosta: Ah, that was a bad copy paste from my end 😅
Just pushed a version with a regular `>` comparison

vsubhuman:
@rooooooooob , check solution plz if you have a chance

rooooooooob (approved):
I guess this is a good enough fix to do it at the end. CIP 2 actually leaves out the fees entirely and says it's out of scope of the CIP so this detail is not covered which is why I tried to do it during the adding phase.

We're also trying to figure out what we want to do for multiassets since those also weren't covered but we can cover that at a later point.

vsubhuman (commented):
/check

vsubhuman:
Thank you, @pedromtcosta ! Look good!

PR #231, opened 2021-10-26
Mint and simpler metadata in tx-builder

vsubhuman:
Added tx-builder functions for simpler metadata and mint

PR #255, opened 2021-11-08
Fix build warnings

pedromtcosta:
I found the 4 types of warnings bellow:
- unnecessary parenthesis on condition (e.g. `if (a == b)`): simply remove the parenthesis;
- several unused imports: simply remove the import
- several unused variables: renamed most of these unused variables to underscore (`_`). This was happening mostly in `match` expressions where the value is just ignored, but there were quite a few cases in implementations of the `deserialize_as_embedded_group` function of the `DeserializeEmbeddedGroup` trait. For the first mentioned cases, it makes perfect sense to just rename the variable, but for the second one, there's the decision to either change the trait (since none of the current implementations are actually using the `len` parameter from the trait) or rename the variable in the implementations. I decided to rename the variables because I am always reluctant to break code contracts. Let me know if someone has arguments for actually changing the `trait`
- function `build_tx_pay_to_multisig` is not being used: ~simply remove the function~ this function is actually a test which was missing the `#[test]` tag


vsubhuman:
@pedromtcosta , resolve conflict please

rust/src/tx_builder.rs
```diff
@@ -593,7 +593,10 @@ impl TransactionBuilder {
                         if potential_pure_above_minimum {
                             new_fee = new_fee.checked_add(&additional_fee)?;
                             change_left = Value::zero();
-                            self.add_output(&TransactionOutput::new(address, &potential_pure_value));
+                            match self.add_output(&TransactionOutput::new(address, &potential_pure_value)) {
+                              Err(e) => panic!(e),
+                              Ok(_) => {}
+                            };
```
vsubhuman, line 599: Because the function `add_change_if_needed` returns a `Result<..., JsError>` - the recommended way to deal with functions like this, that complain about unused return result is to use the `?` unwrapper that will fail-fast to propagate the err in case it's returned by the called function.

e.g.
```
self.add_output(&TransactionOutput::new(address, &potential_pure_value))?;
```

This particular line is also being fixed in a parallel PR:

https://github.com/Emurgo/cardano-serialization-lib/pull/231/files#diff-67896f615f1e60f7474eb6194c0614a005522dc35fb07046855d77260954b972R775

vsubhuman (changes requested):

rust/src/tx_builder.rs
```diff
@@ -2117,56 +2118,6 @@ mod tests {
         assert!(input_total >= Value::new(&tx_builder.min_fee().unwrap().checked_add(&to_bignum(COST)).unwrap()));
     }
 
-    fn build_tx_pay_to_multisig() {
```
vsubhuman, line 2120: I think this tx were supposed to be a test, actually, just missing the `#[test]` tag probably

pedromtcosta: I totally overlooked the assertions at this function 😅

pedromtcosta:
@vsubhuman changes addresses and conflict resolved

vsubhuman (approved):

vsubhuman:
@pedromtcosta , resolve conflicts please and we can merge

pedromtcosta:
@vsubhuman Conflicts resolved

vsubhuman (approved):

PR #256, opened 2021-11-08
Fixing GH-checks to run the build properly + Fixing `rand` dependency wasm support

vsubhuman:


vsubhuman (commented):
/check

vsubhuman (commented):
/check

vsubhuman (commented):
/check

vsubhuman:
After adding the `rand` dependency in https://github.com/Emurgo/cardano-serialization-lib/pull/232 the wasm build started failing with: `error: the wasm32-unknown-unknown target is not supported by default, you may need to enable the "js" feature. For more information see: https://docs.rs/getrandom/#webassembly-support`

image

In **this** PR the gh-checks are fixed to properly run the builds and the build scripts are fixed to properly return the exit status for the checks can fail.

To fix the problem with the `rand` crate I have to add additional explicit dependency:
```
getrandom = { version = "0.2.3", features = ["js"] }
```

vsubhuman (commented):
/check

pedromtcosta (approved):
Looks good! Tested the build locally as well 👍

PR #264, opened 2021-11-21
TxBuilder: CIP2 MultiAsset variants added

rooooooooob:
Implemented an adaption of CIP2LargestFirst to handle multiassets as
the current implementation would only order by ADA resulting in very
poor selections if many inputs were available and non-ADA multiassets
were in the outputs.

The previous CIP2LargestFirst algorithm will now error on multiasset
outputs.

MultiAsset's documentation was improved slightly and convenience methods
`get_asset()` and `set_asset()` are included now to avoid having to
additionally deal with Assets and do it all at once.

RandomImprove was also implemented for multiasset, although we might need to adapt it further or make some changes (see comments in latest commit)


rust/src/tx_builder.rs
```diff
@@ -175,110 +181,244 @@ impl TransactionBuilder {
...
+            .cloned()
+            .collect::<Vec<usize>>();
+        let mut associated_indices: BTreeMap<TransactionOutput, Vec<usize>> = BTreeMap::new();
+        let mut outputs = self.outputs.0.iter()
+            .filter(|output| by(&output.amount).is_some())
+            .cloned()
+            .collect::<Vec<TransactionOutput>>();
+        outputs.sort_by_key(|output| by(&output.amount).expect("filtered above"));
+        for output in outputs.iter().rev() {
+            // TODO: how should we adapt this to inputs being associated when running for other assets?
```
rooooooooob, line 353: I'm not sure how significant this is for most wallets wrt the multiassets. We could do a halfway point and only consider this for the ADA-only stage (e.g. so we count the inputs added to satisfy the multiassets towards the ADA portion). Although it might be not much more work to just count inputs added in previous rounds (for other assets) against the outputs for subsequent rounds in general not just for ADA. It's a bit weird since random-improve is specified to do things per-output rather than for the total output required.

rust/src/tx_builder.rs
```diff
@@ -175,110 +181,244 @@ impl TransactionBuilder {
...
-                            let new_input = available_inputs.get_mut(random_index).unwrap();
-                            let cur = from_bignum(&input.output.amount.coin);
-                            let new = from_bignum(&new_input.output.amount.coin);
-                            let min = from_bignum(&output.amount.coin);
-                            let ideal = 2 * min;
-                            let max = 3 * min;
-                            let move_closer = (ideal as i128 - new as i128).abs() < (ideal as i128 - cur as i128).abs();
-                            let not_exceed_max = new < max;
-                            if move_closer && not_exceed_max {
-                                std::mem::swap(input, new_input);
```
rooooooooob, line 254: I think this part was broken before anyway. Maybe that's why that phase 3 was necessary despite adding in the input fee to the required amount in phase 1 before? The inputs in `associated_inputs` would be distinct copies from the ones in the builder as they are added above as a copy.

rust/pkg/cardano_serialization_lib.js.flow
```diff
@@ -333,6 +343,13 @@ declare export var MetadataJsonSchema: {|
...
  */
 
+declare export var StakeCredKind: {|
+  +Key: 0, // 0
+  +Script: 1, // 1
+|}
+
+/**
+ */
 declare export var CoinSelectionStrategyCIP2: {|
```
rooooooooob, line 353: I messed up the merge here. I'll remove the duplicate `CoinSelectionStrategyCIP2` in the next commit I do since it's a minor change.

rust/src/lib.rs
```diff
@@ -2528,23 +2528,37 @@ impl MultiAsset {
...
+        self.0.get(policy_id).map(|v| v.clone())
+    }
+
+    // sets the asset {asset_name} to {value} under policy {policy_id}
+    // returns the previous amount if it was set, or else None (undefined in JS)
+    pub fn set_asset(&mut self, policy_id: &PolicyID, asset_name: &AssetName, value: BigNum) -> Option<BigNum> {
+        self.0.entry(policy_id.clone()).or_default().insert(asset_name, &value)
+    }
+
+    pub fn get_asset(&self, policy_id: &PolicyID, asset_name: &AssetName) -> BigNum {
```
ozgrakkurt, line 2552: There are no comments for `get_asset` also the comments for `set_asset` aren't doc comments (with triple slashes).
Also is it ok to return `BigNum::zero` instead of None in `get_asset` if the asset doesn't exist?

rust/src/tx_builder.rs
```diff
@@ -265,118 +271,252 @@ impl TransactionBuilder {
...
+                // Phase 3: add extra inputs needed for fees (not covered by CIP-2)
+                // We do this at the end because this new inputs won't be associated with
+                // a specific output, so the improvement algorithm we do above does not apply here.
+                while input_total.coin < output_total.coin {
+                    if available_indices.is_empty() {
+                        return Err(JsError::from_str("UTxO Balance Insufficient[x]"));
                     }
-                    // differing from CIP2, we include the needed fees in the targets instead of just output values
+                    let i = *available_indices.iter().nth(rng.gen_range(0..available_indices.len())).unwrap();
+                    available_indices.remove(&i);
```
ozgrakkurt, line 314: Using a Vec instead of BTreeSet is better for performance here. Because you select a random element and selecting random elements from a Vec is constant time. You can move the last element to the removed element's place when removing an element if you use a Vec.

rust/src/tx_builder.rs
```diff
@@ -265,118 +271,252 @@ impl TransactionBuilder {
...
-                    let mut added = Value::new(&Coin::zero());
-                    let mut needed = output.amount.clone();
-                    while added < needed {
-                        if available_inputs.is_empty() {
-                            return Err(JsError::from_str("UTxO Balance Insufficient"));
+            CoinSelectionStrategyCIP2::LargestFirstMultiAsset => {
+                // indices into {available_inputs} for inputs that contain {policy_id}:{asset_name}
+                let mut available_indices = (0..available_inputs.len()).collect::<Vec<usize>>();
+                // run largest-fist by each asset type
+                if let Some(ma) = output_total.multiasset.clone() {
```
ozgrakkurt, line 326: I think clone might no be needed here since we only iterate over the data

rust/src/tx_builder.rs
```diff
@@ -265,118 +271,252 @@ impl TransactionBuilder {
...
+        // iterate in decreasing order for predicate {by}
+        for i in relevant_indices.iter().rev() {
+            if by(input_total).unwrap_or(BigNum::zero()) >= by(output_total).expect("do not call on asset types that aren't in the output") {
+                break;
+            }
+            let input = &available_inputs[*i];
+            // differing from CIP2, we include the needed fees in the targets instead of just output values
+            let input_fee = self.fee_for_input(&input.output.address, &input.input, &input.output.amount)?;
+            self.add_input(&input.output.address, &input.input, &input.output.amount);
+            *input_total = input_total.checked_add(&input.output.amount)?;
```
ozgrakkurt, line 416: Doing `?` here might not return a sensible error since that would be `JsError::from(None)`

rust/src/tx_builder.rs
rooooooooob: You have to look at where it's used within `random_improve_by()` too where it's passed into. One thing to remember there is that you're indexing it by value not by index so you'd have an O(n) look-up (e.g. for `position()` even if the `swap_remove()` call itself is O(1). You can take out the parameter to `nth()` call here in Phase 3 but in Phase 1/2 we can't re-use the `random_index` down there (in `random_improve_by()`) either to pass into `swap_remove()` because that's an index into `relevant_indices` not `available_indices` so the indices won't have anything in common - but the values inside them will. I actually had `available_indices` as a Vec like in largest-first before changing it. Phase 3 is also likely to be run very few times relative to Phase 1/2 as it is only to cover the difference in fees caused by the inputs added in phases 1/2. The bulk of the ADA should be added beforehand. You're choosing between O(n) vs O(logn) possibly a bunch of times in phase 1/2 followed by O(1) vs O(n) maybe once or twice in phase 3.

There are, however, differences in cache/locality/allocations although BTreeMap has contiguous sections within it so I'm not sure if it is even much of an issue in practice vs Vec. I have no idea how all this translates into wasm though which is where most of our users are using it from. It's possible this could be significant relative to algorithmic differences but until we start to see it being an issue I'm not sure how much we should care.

rust/src/tx_builder.rs
rooooooooob: We plan on going through the builder in the coming while to improve errors so they're more helpful for the user so it's good you're looking at these. Where does `None` come from though? I thought it would return the error type from `BigNum::checked_add()` (that seems to be the only part where `Value::checked_add()` can error) which is `JsError::from_str("overflow")`.

rust/src/tx_builder.rs
rooooooooob: It was to placate the borrow checker since we pass in `output_total` by mutable borrow into the function which wouldn't be possible if we had an immutable borrow in that if above. It doesn't want there to be problems with iterator invalidation even if in this case we're only modifying the `coin` wrapped-u64 within `output_total`, but who knows what could happen if we happened to modify the `multiasset` parts within there with (de)allocatations.... I'm definitely not having any C++ flashbacks here so thanks borrow checker :)

If it becomes a performance issue we could pass the multi-asset part by immutable ref and the coin portion as mutable ref separately and then have to refactor the compare code used by Value to take in both by ref separately instead of as a whole Value, but that sounds like a lot of hassle and I don't think one clone for just the sum of outputs is going to make much of a difference at all.

rust/src/lib.rs
rooooooooob: Good catch, I pushed a commit to update the comments.

The first time I wrote `get_asset()` I had it returning `Option<BigNum>` but it felt more awkward to use when I was writing the tests so I changed it to just `BigNum`. Does anyone else have any opinion on it? It seems the same thing to me conceptually.

rust/src/tx_builder.rs
ozgrakkurt: Thanks for the explanation. I agree this should be good than

rust/src/tx_builder.rs
ozgrakkurt: Oh, I assumed it was an integer or float type. Sorry about that, I realize it is already a JsError when returned from checked_add.

rust/src/lib.rs
```diff
@@ -2528,23 +2528,40 @@ impl MultiAsset {
...
+    /// the number of unique policy IDs in the multiasset
     pub fn len(&self) -> usize {
         self.0.len()
     }
 
-    pub fn insert(&mut self, key: &PolicyID, value: &Assets) -> Option<Assets> {
-        self.0.insert(key.clone(), value.clone())
+    /// set (and replace if it exists) all assets with policy {policy_id} to a copy of {assets}
+    pub fn insert(&mut self, policy_id: &PolicyID, assets: &Assets) -> Option<Assets> {
+        self.0.insert(policy_id.clone(), assets.clone())
```
NicolasDP, line 2538: this is something I'm seeing everywhere in the rust library, why are we passing values by reference if it is to clone them. Why not make the API take by values and let the user make the right choices with the memory ?

rust/src/tx_builder.rs
```diff
@@ -265,118 +271,252 @@ impl TransactionBuilder {
     /// This function, diverging from CIP2, takes into account fees and will attempt to add additional
     /// inputs to cover the minimum fees. This does not, however, set the txbuilder's fee.
     pub fn add_inputs_from(&mut self, inputs: &TransactionUnspentOutputs, strategy: CoinSelectionStrategyCIP2) -> Result<(), JsError> {
-        let mut available_inputs = inputs.0.clone();
+        let available_inputs = &inputs.0.clone();
```
NicolasDP, line 274: why do you clone to take a reference of the array ?

rust/src/lib.rs
rooooooooob: It's for when it's consumed on the JS side via wasm, which is most of our users. Move semantics don't translate very well over to JS/wasm and there's no rust borrow checker to help you over there. If it were a rust-only library I would 100% agree though.

ozgrakkurt: I think maybe we should convert all rust code to idiomatic rust code and make a js api over it so it is good to use on js side.

NicolasDP: I see. Well this is a problem inherent to the whole rust library here. so not necessarily something to do in this P. Though I thought I'd just ask the reason and see what the answer would be.

rooooooooob: I'm not sure how well we could simply "make a js api over it".

If we go about it by wrappers on the JS side which contain an object in wasm bound by wasm-bindgen we have these cons:
- who owns the memory? how do we track that in JS? (maybe this is solvable but afaik there's no RAII/destructor we could rely on here to ref-count or whatever) simply putting it in a Rc/Arc wrapper in rust just shifts the problem since the js object that points to that itself could be copied/go out of scope/etc with no way (afaik - I could be wrong and I hope I am!) for rust to be aware of that. Maybe `js_sys` has something could help there though, I haven't looked much.

If we go about it by having all the data be on the JS side natively then we have the following cons:
- we would have to rewrite so much of the lib in JS
- we would have to write conversion code to go to/from JS to wasm for all types
- this might have a significant performance penalty going over the boundary multiple times instead of just relying on ptrs to wasm linear memory where the objects exists like now.

If we go about it by having there be idiomatic rust code with separate rust wasm-bindgen annotated newtypes ([js-chain libs reference](https://github.com/input-output-hk/js-chain-libs/blob/master/src/lib.rs). This is what IOHK did for but this made sense since the rust code was already written and they were just providing a wasm API around it) we have these cons:
- experience is equally awful from JS (possibly slightly worse performance since more indirection but likely irrelevant), but it's better from rust. keep in mind most users are using the js bindings not the rust crate afaik. the npm packages have over 50x the weekly downloads as the rust crate - maybe not the best since performance matters to some projects which could be in rust and be more important than many smaller js users.

All of these also mean doc-comments are harder to auto-gen. Naively doing these wrappers by hand would mean we'd have to copy them manually (instead of wasm-bindgen exposing them automatically) which means they could get out of date easily if they're in two places. If we're doing this via a wasm-bindgen annotated wrapper in rust we could  produce whatever wrapper by a macro we could probably extract them but then we're stuck with very 1:1 or formulaic bindings.

But yeah this probably isn't the place to discuss this. We've had a discussion (sparked by this PR) internally in our slack but maybe we want to discuss it in a public issue here instead.

ozgrakkurt: I am opening a public issue for this

vsubhuman (approved):

PR #270, opened 2021-11-24
change TransactionBuilderConfigBuilder API for js usage

ozgrakkurt:
This change to TransactionBuilderConfig makes it so it is easier and less error prone when using from JavaScript.
See #269 for discussion.


vsubhuman (approved):

vsubhuman:
Good call making the clone before setting the value in setters, I got that wrong last time.

Issue #37, opened 2020-07-29
Wrapper for TransactionBody to use for signing


SebastienGllmt:
Currently when you call `build` on the TransactionBuilder, you get back a `TransactionBody`. This isn't great because you lose

1) Information about which witnesses are needed
2) Information about the "amount" of each input

It'd be better if we made the `build` function instead give you a `TransactionSigner` with functions to add witnesses, and its give you an error if you're missing a witness at the end (instead of being told the witness is missing once you submit it to your Cardano node)

SebastienGllmt:
I think if we have this, we should probably also do the following at the same time

1. Get rid of `get_fee_or_calc` function entirely
2. Rename `estimate_fee` to `min_fee`
3. Rename `add_change_if_needed` to `close_with_change` and return a TransactionSigner
4. Add a `close_and_burn` which returns a TransactionSigner

PR #273, opened 2021-11-28
Fixing minting API

vsubhuman:
Reworked mint helper function to accept minting script instead of just the hash, and added storing these scripts in the resulting witness set

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -34,6 +34,7 @@ use cbor_event::{
     se::{Serialize, Serializer},
 };
 
+pub mod traits;
 pub mod address;
 pub mod chain_core;
 pub mod chain_crypto;
@@ -59,6 +60,7 @@ use metadata::*;
 use utils::*;
 use std::cmp::Ordering;
 use std::collections::BTreeSet;
+use crate::traits::NoneOrEmpty;
 
 type DeltaCoin = Int;
 
@@ -1732,6 +1734,21 @@ impl NativeScripts {
 
     pub fn add(&mut self, elem: &NativeScript) {
         self.0.push(elem.clone());
+    }
+}
+
+impl From<Vec<NativeScript>> for NativeScripts {
+    fn from(scripts: Vec<NativeScript>) -> Self {
+        scripts.iter().fold(NativeScripts::new(), |mut scripts, s| {
+            scripts.add(s);
+            scripts
+        })
+    }
+}
+
+impl NoneOrEmpty for NativeScripts {
+    fn is_none_or_empty(&self) -> bool {
+        self.0.is_empty()
     }
 }
 
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -224,6 +224,7 @@ impl Certificates {
 }
 
 pub type RequiredSigners = Ed25519KeyHashes;
+pub type RequiredSignersSet = BTreeSet<Ed25519KeyHash>;
 
 #[wasm_bindgen]
 #[derive(Clone)]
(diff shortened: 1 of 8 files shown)


rust/src/tx_builder.rs
```diff
@@ -638,29 +655,66 @@ impl TransactionBuilder {
 
     /// Set explicit Mint object to this builder
     /// it will replace any previously existing mint
+    /// NOTE! If you use `set_mint` manually - you must use `set_mint_scripts`
+    /// to provide matching policy scripts or min-fee calculation will be rejected!
```
rooooooooob, line 659: Is it possible to enforce this is set if you set the mint in the build function and return an error otherwise?

vsubhuman: 🤔 Might be a good idea. Need to think a bit, because it's inconsequential for the original `build()` as it only returns the tx body. Might be added into the `build_tx()`, but on the other hand the only place where it actually breaks the logic is the fee estimation as we can guarantee that the estimate will be wrong without the scripts being present. That's why I have already added the assertion and an error into the `min_fee` function in case matching scripts are not found.

vsubhuman: @rooooooooob , what do you think of combining the `set_min` and `set_mint_scripts` into a single function like

```
/// Set explicit Mint object and witness mint NativeScripts to this builder
/// it will replace any previously existing mint and witness scripts
/// NOTE! The scripts must contain all entries to match policy hashes used in the mint
/// If no matching script is found for a policy - error will be returned and the builder state will not change
pub fn set_mint(&mut self, mint: &Mint, mint_scripts: &NativeScripts) -> Result<(), JsError> { ... }
```

ozgrakkurt (commented):
looks good to me

rust/src/tx_builder.rs
rooooooooob: That sounds like an even better idea since it's simpler and they're gonna need it to make the full tx later anyway.

vsubhuman: Will add the change for this then

vsubhuman (commented):
/check

vsubhuman:
@rooooooooob @ozgrakkurt , plz review, this one is ready now

vsubhuman (commented):
/check

rust/src/tx_builder.rs
```diff
@@ -1077,13 +1142,21 @@ impl TransactionBuilder {
         }
     }
 
+    fn get_witness_set(&self) -> TransactionWitnessSet {
```
rooooooooob, line 1145: We should be very explicit about what this includes and what doesn't. If you didn't look at the code you might not know it's only talking about the native scripts at all and might expect other witnesses. We have an issue #37 which would be nice to implement one day to help people sign txs after but as-is we don't have any of the info available to construct the other witnesses so unless we're very explicit about this function name (or at the very least documentation) I think it can be misleading.

rooooooooob (commented):
Looks good for the most part. I just think we should better document (or change the name of) `TransactionBuilder::get_witness_set()`

rust/src/tx_builder.rs
```diff
@@ -140,7 +143,33 @@ fn fake_full_tx(tx_builder: &TransactionBuilder, body: TransactionBody) -> Resul
...
 
+fn assert_required_mint_scripts(mint: &Mint, maybe_mint_scripts: Option<&NativeScripts>) -> Result<(), JsError> {
+    if maybe_mint_scripts.is_none_or_empty() {
+        return Err(JsError::from_str(
+            "Mint is present in the builder, but witness scripts are not provided!",
+        ));
+    }
+    let mint_scripts = maybe_mint_scripts.unwrap();
+    let witness_hashes: HashSet<ScriptHash> = mint_scripts.0.iter().map(|script| {
+        script.hash(ScriptHashNamespace::NativeScript)
```
rooooooooob, line 154: It's possible this will be a bottleneck as `fee_for_input()` (which in turn calls `min_fee()` and ultimately this) is called over and over when adding inputs or considering it.

Of course we can always profile this later as part of an entire typical workflow to actually see since this is just my guess based on the last time I profiled the builder, which didn't use automatic input selection. I'm not 100% sure if the hashing was as bad as the key conversion operations either. And if it is a bottleneck we can just make a new PR since it won't even affect major version number or anything. Avoiding it might make code complicated too so I think it's totally fine to just merge this PR with the current behavior. I just wanted to document this now in this review so I don't forget later.

rust/src/tx_builder.rs
SebastienGllmt: yeah this is a long-standing issue so I got the ball rolling by creating https://github.com/Emurgo/cardano-serialization-lib/pull/307

rooooooooob: We'll still want to document this even with #307 or will we restrict the witness builder API so it's a non-issue? (e.g. can't set native scripts after-the-fact so it must be done via the tx builder - or would we not want that?)

rooooooooob:
@vsubhuman what's the status on this PR? It would be nice to get it merged if we can so we can not conflict with other TxBuilder changes, plus I started updating the documentation and wanted to be able to include these changes in that.

vsubhuman:
> @vsubhuman what's the status on this PR? It would be nice to get it merged if we can so we can not conflict with other TxBuilder changes, plus I started updating the documentation and wanted to be able to include these changes in that.

I would be really happy to get that in and can prioritize it this week. Let me look into the comments and see if I can make any required changes quickly. And then we can get the coin-selection in as well

rooooooooob:
That would be great. My comment about documenting what is in the returned witness set isn't quite as important anyway since it looks like we'll be working to integrate in a second-phase witness builder combining this + #307 and hooking the two together later on, so that function return will likely change soon anyway.

rust/src/lib.rs
```diff
@@ -221,6 +224,7 @@ impl Certificates {
 }
 
 pub type RequiredSigners = Ed25519KeyHashes;
+pub type RequiredSignersSet = BTreeSet<Ed25519KeyHash>;
```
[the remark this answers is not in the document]
vsubhuman: Afaik, order does not matter at all, but at the same time not sure if should be forced to always be unique. Careful rework to `RequiredSigners` also be a set internally can be done later probably, don't want to delay v10 more for this.

vsubhuman:
> Looks good for the most part. I just think we should better document (or change the name of) `TransactionBuilder::get_witness_set()`

Added comment to this function. Note that it's internal, so not a part of the public API, but good point that it was missing any clarity. Idea is that this function should be updated when new functionality is added to the builder, e.g. when it supports more different scripts.

rust/src/tx_builder.rs
```diff
@@ -274,7 +303,7 @@ pub struct TransactionBuilder {
     validity_start_interval: Option<Slot>,
     input_types: MockWitnessSet,
     mint: Option<Mint>,
-    inputs_auto_added: bool,
```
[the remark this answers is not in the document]
vsubhuman: I couldn't think of anything for now. Can be added back once we have a need for it.

rust/src/tx_builder.rs
```diff
@@ -650,31 +679,65 @@ impl TransactionBuilder {
...
+        let mint_scripts = {
+            let mut witness_scripts = self.mint_scripts.as_ref().cloned()
+                .unwrap_or(NativeScripts::new());
+            if is_new_policy {
+                // If policy has not been encountered before - insert the script into witnesses
+                witness_scripts.add(&policy_script.clone());
+            }
+            witness_scripts
+        };
+        self.mint = Some(mint.clone());
```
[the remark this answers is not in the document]
vsubhuman: Good catch, removed it

rust/src/tx_builder.rs
vsubhuman: Good point. I have commented out the check in `min_fee` for now as it is not critical and was mostly added for extra assurance. Added a comment there that it can be reworked to cache hashing or maybe only perform the check once if the mint has not changed since, but not enough time to work on this atm

vsubhuman (commented):
/check

vsubhuman (commented):
/check

rooooooooob (approved):
LGTM now

vsubhuman (commented):
/check

Issue #276, opened 2021-12-02
Current state of rust library API and implementation


ozgrakkurt:
Currently rust api and implementation of the library isn't idiomatic rust code, which makes it awkward to read and understand.

I think we can change the library so it is like a normal rust library, but wrap it in javascript so It is convenient to use from javascript as well. I am currently working on a concrete proposal/PR to do this.

Also currently running clippy fails on the rust library and running fmt changes so much lines that it is hard to apply to the whole library and not break a lot of people's work.

See also:
- [https://github.com/Emurgo/cardano-serialization-lib/pull/264#discussion_r759795133](https://github.com/Emurgo/cardano-serialization-lib/pull/264#discussion_r759795133)
- #270 
- #269

rooooooooob:
>I am currently working on a concrete proposal/PR to do this.

Could you provide some details about this please @ozgrakkurt ?

After writing those comments in #264 I started trying something exploratory so that we can better weigh the options with what I mentioned in #264's comment about using a macro. I'm trying to see what could be done via procedural macros to make wrappers similar to the wrappers I linked in js-chain-libs. My hope is that it could even do parameter and error conversions but I'm a noob at proc macros so we'll have to see how it goes.

I don't want to commit to any approach at this moment but I wanted to explore this option at least to give us a better idea. This would help the usage from rust as mentioned but the usage from js would still be the same as it is now (i.e. not great).

Also I'm linking these wasm-bindgen issues here since they're highly relevant to making the JS experience better:

rustwasm/wasm-bindgen#111
rustwasm/wasm-bindgen#1742

We need to remember that the npm package users are far, far more numerous than the rust crate users as well.

rooooooooob:
@ozgrakkurt I was thinking about this more. We're going to need *some* kind of wasm_bindgen interface no matter what, whether that be auto-conversions applied programatically, or whether it be hand-written wrappers (e.g. how js-chain-libs or Ergo's sigma-rust wasm bindings work), or whether we use code-gen to do it for us.

[...]

Issue #213, opened 2021-09-26
[performance] fake_full_tx in TransactionBuilder should use static data instead of cryptography


CantTouchDis:
I have been playing around building transaction with the TransactionBuilder (version `8.1.0` and now `9.1.0`) and stumbled across a problem with `add_change_if_needed`. For some reason this function takes a very long time in my browser. (brave)

I am not a webdev and just wanted to try out if I could get it to work, so it might very well be a problem with my code.

```js
  const txBuilder = S.TransactionBuilder.new(
    // all of these are taken from the mainnet genesis settings
    // linear fee parameters (a*size + b)
    S.LinearFee.new(S.BigNum.from_str('44'), S.BigNum.from_str('155381')),
    // minimum utxo value
    S.BigNum.from_str('1000000'),
    // pool deposit
    S.BigNum.from_str('500000000'),
    // key deposit
    S.BigNum.from_str('2000000'),
    // max_value_size
    300,
    // max_tx_size
    8000
  );
  // add the input
  txBuilder.add_input(
    bestUtxo.output().address(),
    bestUtxo.input(),
    bestUtxo.output().amount()
  );
  // add the input
  txBuilder.add_output(...);
  // calculate change (THIS TAKES FOR AGES)
  txBuilder.add_change_if_needed(
    changeAddr
  );
  var body = txBuilder.build();

  var transaction = S.Transaction.new(
    body,
    S.TransactionWitnessSet.new(),
    // auxData
  );
  ... sign, set witness and submit
```

Info: I did all of this with testnet addresses, have not tried it yet with mainnet ones. `bestUtxo` in the above is just one `TransactionUnspentOutput` with enough ada, returned by nami wallet

Ignoring the parameters `max_value_size` and `max_tx_size` as it happened in the previous version for me as well, what could be the issue?

Callgraph profiling in my browser:

![ChangeIfNeeded](https://user-images.githubusercontent.com/5838555/134809806-81f1d7b3-17b2-436a-b87d-12140dfabd40.png)

If you need any more information please let me know.

CantTouchDis:
With the 9.1.0 lib I got readable names in the Profiler view:
![Profile9 1 0](https://user-images.githubusercontent.com/5838555/134814798-9d3d786e-b260-4a21-90f7-6ff4031fd7db.png)

CantTouchDis:
I just tried it in my firefox and got the same result for this minimal example code: (assuming S is the loaded module)
```js
[...]

PR #286, opened 2021-12-10
Remove all cryptoxide calls from fake_full_tx()

rooooooooob:
Profiling the tx builder showed that even after we removed the
fake_key_root construction in #214 due to terrible performance in asm.js
(#213) that the cryptoxide calls within `fake_full_tx()` were up a
ridiculous amount of the runtime. I investigated this as we had a report
that it was taking up to hundreds of milliseconds (in rust) just to make a tx and
I wanted to see if there would be any significant improvement by
migrating to an idiomatic rust API as discussed in #276 but from a
purely performance perspective in a release build (opt-level=3) this seemed to
be pretty minor, even with the unnecessary cloning forced by this
inside of the implementation.

This will need to be slightly modified once #273 gets merged as there are
some conflicts.

Previous perf results from building 100,000 simple tx bodies (no witnesses) and serializing
them:
```
-   53.64%        [.] <&cryptoxide::curve25519::Fe as
core::ops::arith::Mul>::mul
    - 53.63% 0xffffffffffffffff
      + 23.03% cardano_serialization_lib::tx_builder::min_fee
      - 21.32%
cardano_serialization_lib::tx_builder::TransactionBuilder::add_change_if_needed
          - 15.21%
cardano_serialization_lib::tx_builder::TransactionBuilder::fee_for_output
              cardano_serialization_lib::tx_builder::min_fee
            + cardano_serialization_lib::tx_builder::fake_full_tx
          + 6.11%
cardano_serialization_lib::tx_builder::TransactionBuilder::min_fee
(inlined)
      - 7.78% ser_lib_perf::main
          + 6.11%
cardano_serialization_lib::tx_builder::TransactionBuilder::build
          + 1.67%
cardano_serialization_lib::tx_builder::TransactionBuilder::add_change_if_needed
      + 1.46% std::rt::lang_start_internal
-   19.40%        [.] cryptoxide::curve25519::Fe::square
    - 0xffffffffffffffff
      - 8.34% cardano_serialization_lib::tx_builder::min_fee
            cardano_serialization_lib::tx_builder::TransactionBuilder::build
            cardano_serialization_lib::tx_builder::TransactionBuilder::build_and_size
          + cardano_serialization_lib::tx_builder::fake_full_tx
      - 7.41%
cardano_serialization_lib::tx_builder::TransactionBuilder::add_change_if_needed
          + 5.58%
cardano_serialization_lib::tx_builder::TransactionBuilder::fee_for_output
          + 1.83%
cardano_serialization_lib::tx_builder::TransactionBuilder::min_fee
(inlined)
      + 2.77% ser_lib_perf::main
      + 0.88% std::rt::lang_start_internal
+    7.80%        [.] cryptoxide::curve25519::GePrecomp::maybe_set
+    5.13%        [.] cryptoxide::curve25519::Fe::invert
+    3.98%        [.] <&cryptoxide::curve25519::GeP3 as
core::ops::arith::Add<&cryptoxide::curve25519::GePrecomp>>::add
+    2.84%        [.] cryptoxide::curve25519::GePrecomp::select
+    1.22%        [.] cryptoxide::curve25519::ge_scalarmult_base
+    1.00%        [.]
cryptoxide::sha2::impl512::reference::digest_block_u64
      0.47%        [.] cryptoxide::curve25519::GeP2::dbl
      0.13%        [.] cryptoxide::curve25519::sc_muladd
      0.12%        [.] cryptoxide::curve25519::Fe::to_bytes
      0.12%        [.] cryptoxide::curve25519::sc_reduce
      0.10%        [.] cbor_event::se::Serializer<W>::write_type
      0.07%        [.]
cardano_serialization_lib::tx_builder::fake_full_tx
```

which shows that almost the entire runtime was spent on cryptoxide calls
which were only necessary for `fake_full_tx()`.

Running it again afterwards there is no overwhelming bottleneck anymore
and the remaining runtime was fairly distributed.

### DIFF
--- a/rust/src/tx_builder.rs
+++ b/rust/src/tx_builder.rs
@@ -47,10 +47,23 @@ fn fake_private_key() -> Bip32PrivateKey {
 }
 
 fn fake_key_hash() -> Ed25519KeyHash {
-    Ed25519KeyHash::from_bytes(
-        vec![142, 239, 181, 120, 142, 135, 19, 200, 68, 223, 211, 43, 46, 145, 222, 30, 48, 159, 239, 255, 213, 85, 248, 39, 204, 158, 225, 100]
+    Ed25519KeyHash::from(
+        [142, 239, 181, 120, 142, 135, 19, 200, 68, 223, 211, 43, 46, 145, 222, 30, 48, 159, 239, 255, 213, 85, 248, 39, 204, 158, 225, 100]
+    )
+}
+
+fn fake_raw_key_sig() -> Ed25519Signature {
+    Ed25519Signature::from_bytes(
+        vec![36, 248, 153, 211, 155, 23, 253, 93, 102, 193, 146, 196, 181, 13, 52, 62, 66, 247, 35, 91, 48, 80, 76, 138, 231, 97, 159, 147, 200, 40, 220, 109, 206, 69, 104, 221, 105, 23, 124, 85, 24, 40, 73, 45, 119, 122, 103, 39, 253, 102, 194, 251, 204, 189, 168, 194, 174, 237, 146, 3, 44, 153, 121, 10]
     ).unwrap()
 }
+
+fn fake_raw_key_public() -> PublicKey {
+    PublicKey::from_bytes(
+        &[207, 118, 57, 154, 33, 13, 232, 114, 14, 159, 168, 148, 228, 94, 65, 226, 154, 181, 37, 227, 11, 196, 2, 128, 28, 7, 98, 80, 209, 88, 91, 205]
+    ).unwrap()
+}
+
 
 // tx_body must be the result of building from tx_builder
 // constructs the rest of the Transaction using fake witness data of the correct length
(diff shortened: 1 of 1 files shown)


rooooooooob:
I should note that those inner calls when I expanded the top few are inclusive of the underlying cryptoxide calls. The cryptoxide calls were something like 95%+ of the runtime. For exclusive samples not a single non-cryptoxide function took up more than a fraction of a percent of total samples.

This is also only the time for the transaction body creation (full `TransactionBuilder` flow + also serializing the resulting tx body and summing the total lengths of the generated CBOR to print out at the end so I could assure myself that the tx building or any other part wasn't being optimized out since this was a release build (with debug symbols) that I was profiling). This seems to line up with the massive difference in regular release builds (albeit in asm.js but I bet the difference is there in other compilation targets but they're just faster in general so no one notices if they're only building 1 tx) when someone removed the other crpyto code (I assume it was the worst offender) in #214

vsubhuman (approved):

Issue #281, opened 2021-12-06
[BETA-8] Can't seem to find a way to attach datum and redeemer besides witnessSet


Mercurial:
Hi guys I'm trying to submit a plutus/alonzo(sc) transaction consuming a input from a validator, I'm trying to match the cli functionality but I just cant seem to find a way to attach the datum / redeemer besides the witnessSet. Can someone show me how to use beta-8 to submit a plutus/alonzo transaction? 

I'm contructing the tx manually since txBuilder doesn't support calculating / running plutus interpreter "yet"

Here is my current code: 

```ts
const transactionWitnessSet = Cardano.TransactionWitnessSet.new();
const transactionInputs = Cardano.TransactionInputs.new();
const transactionOutputs = Cardano.TransactionOutputs.new();

const baseAddress = Cardano.BaseAddress.from_address(
    Cardano.Address.from_bech32("addr_test1qqly5tk8pl80ne2vgdapwdc568uzh93yyducz6l28hfc0htkdqgkydtswrycyf9hruerftc8mwel9ck6pksvyszs968qhauh2s")
) as BaseAddress;

const requiredSigners = Cardano.Ed25519KeyHashes.new();
requiredSigners.add(baseAddress.payment_cred().to_keyhash() as Ed25519KeyHash);

const plutusDatum = HoskySwapDatum() as PlutusData;
const datumList = Cardano.PlutusList.new();
datumList.add(plutusDatum);

const redeemers = Cardano.Redeemers.new();
redeemers.add(SimpleRedeemer() as Redeemer);

transactionWitnessSet.set_plutus_scripts(ContractScript() as PlutusScripts);
transactionWitnessSet.set_plutus_data(Cardano.PlutusList.from_bytes(datumList.to_bytes()));
transactionWitnessSet.set_redeemers(Cardano.Redeemers.from_bytes(redeemers.to_bytes()));

transactionInputs.add(
    Cardano.TransactionInput.new(
        Cardano.TransactionHash.from_bytes(fromHex("ca240a8fa7ef1a86662b1ad0064566f76808748cd2c1a83a4041beac1c5a4efc")), 0
    )
);

transactionInputs.add(
    Cardano.TransactionInput.new(
        Cardano.TransactionHash.from_bytes(fromHex("63461348d5b18a4f05d28abaf4930d160912d5ac99433f4487d67c704f5b8fdf")), 0
    )
);

transactionOutputs.add(
    Cardano.TransactionOutput.new(
[...]

PR #287, opened 2021-12-10
Added plutus default cost-model to constants

vsubhuman:
Motivated by this issue: https://github.com/Emurgo/cardano-serialization-lib/issues/281#issuecomment-988313923

### DIFF
--- a/rust/src/constants.rs
+++ b/rust/src/constants.rs
@@ -1,0 +1,52 @@
+use super::*;
+use crate::Epoch;
+use crate::plutus::{Costmdls, CostModel, Language};
+
+// A list of entries where the first element is the epoch after which the cost-model is the correct one to be used.
+// The second element is the cost model, which is an array of 166 operations costs, ordered by asc operaion names.
+// The third value is the pre-calculated `language_views_encoding` value required for the script hash creation.
+// The cost-model values are taken from the genesis block - https://github.com/input-output-hk/cardano-node/blob/master/configuration/cardano/mainnet-alonzo-genesis.json#L26-L195
+// The keys on the genesis block object (operation names) are sorted plain alphabetically.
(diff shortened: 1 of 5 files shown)


vsubhuman:
```
hash_script_data(
  &redeemers,
  &Constants::plutus_default_cost_models(<current_epoch>),
  datums,
)
```
or
```
hash_script_data2(
  &redeemers,
  &Constants::plutus_default_language_views_encoding(<current_epoch>),
  datums,
)
```

vsubhuman (commented):
/check

rust/src/constants.rs
```diff
@@ -0,0 +1,58 @@
...
+
+#[wasm_bindgen]
+pub struct Constants();
+
+#[wasm_bindgen]
+impl Constants {
+
+    /// The function accepts the number of the "current" epoch and returns the
+    /// default Plutus cost-models object to be used in that epoch
+    pub fn plutus_default_cost_models(_epoch: Epoch) -> Costmdls {
```
rooooooooob, line 26: Aren't these also just protocol parameters that can be changed just as arbitrarily as others that we are trying to maintain not hard coded? (e.g. bytes per utxo word, pool deposit etc that are parameters for the tx builder).

It seems inconsistent to provide constants in this case but not there.

And since this is a static library how would we use that `_epoch` parameter to change anything? In what situations would we have multiple different responses here? I could be missing something important since I don't know much about Plutus but it seems like you'd always want the most up to date ones so that you can submit the tx successfully. Is there a use case for historical verification that we would just need this for instead of maintaining a whole running ledger? And as for future data, can we do anything with that parameter? I see `max_epoch` in `ProtocolParamUpdate` but I'm not sure if that means it's invalid after that or if that's more like the ttl on txs. I don't know all the inner workings of Cardano updates so clarifications here would be appreciated.

PR #290, opened 2021-12-12
added min ADA calculation for more precise max value

alessandrokonrad:
I had a few cases, where the max value size was exceeded by a few bytes, because the add_change_if_needed function doesn't take ADA into consideration when it splits the utxos, so I added the minADA to make the calculation more precise.

### DIFF
--- a/rust/src/tx_builder.rs
+++ b/rust/src/tx_builder.rs
@@ -822,7 +822,7 @@ impl TransactionBuilder {
                 }
                 let change_estimator = input_total.checked_sub(&output_total)?;
                 if has_assets(change_estimator.multiasset()) {
-                    fn will_adding_asset_make_output_overflow(output: &TransactionOutput, current_assets: &Assets, asset_to_add: (PolicyID, AssetName, BigNum), max_value_size: u32) -> bool {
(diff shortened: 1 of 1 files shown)


rust/src/tx_builder.rs
```diff
@@ -878,7 +882,7 @@ impl TransactionBuilder {
                                 let asset_name = asset_names.get(n);
                                 let value = assets.get(&asset_name).unwrap();
 
-                                if will_adding_asset_make_output_overflow(&output, &rebuilt_assets, (policy.clone(), asset_name.clone(), value), max_value_size) {
+                                if will_adding_asset_make_output_overflow(&output, &rebuilt_assets, (policy.clone(), asset_name.clone(), value), max_value_size, coins_per_utxo_word) {
```
rooooooooob, line 885: Note: #304 updates this so you might need to fix that before we can merge this depending on which one gets merged first

rooooooooob (approved):
LGTM, nice catch

vsubhuman (approved):
Thank you, @alessandrokonrad !

vsubhuman (commented):
/check

PR #291, opened 2021-12-13
run cargo fmt on rust code

ozgrakkurt:
Run cargo fmt on entire rust code. This should be ok since:
- Other branches can run `cargo fmt` as well and get rid of diffs.
- Running fmt on generated code is ok since it is generated very infrequently.


vsubhuman (commented):
/check

Issue #263, opened 2021-11-19
Possibly confusing binary formats


rooooooooob:
For **most** types we have the `to_bytes()` and `from_bytes()` formats using the direct CBOR format in the [alonzo.cddl binary spec](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/alonzo/test-suite/cddl-files/alonzo.cddl).

This causes some confusion for various types that are essentially defined as just `bytes` (possibly with size restrictions). These types are mostly crypto related and consist of all the hash (`Ed25519KeyHash`, `TransactionHash` etc) and key (`PublicKey`, `Bip32PublicKey`, etc) types as well as `Address`. I believe these are the only ones but there could be more.

These types already had well-defined binary formats, whereas others simply had those `to_bytes()`/`from_bytes()` blanket-implemented off the CBOR format and the concern was just being self-consistent primarily, not with having every single `to_bytes()`/`from_bytes()` all be valid CBOR or not.

It wasn't as big of an issue worth breaking the API over before but now with increasing use-cases directly building off of the CDDL format it has already become a source of trouble and confusion. One example is how the [CIP-30 dApp connector](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0030) defines most types as hex-encoded CBOR based off of alonzo.cddl for types used directly in the API. This poses an issue with `Address` in particular where according to the spec, for consistency against alonzo.cddl (since we don't want to tie any CIP to a specific library) you would not be following the spec if you use our `Address::to_bytes()`/`Address::from_bytes()`.

While this could be a local fix for just `Address`, I'm sure this will come up again somewhere else and it just is a source of slight confusion waiting to catch random unsuspecting users.

My first thoughts are to do the following three things:
1. keep `to_bytes()`/`from_bytes()` as is everywhere and document that this is simply a self-consistent (e.g. `A::from_bytes(a.to_bytes())` should run correctly and return essentially a copy of `a`) format and may or may not be CBOR bytes. These could also possibly be marked as deprecated.
2. add in `to_raw_bytes()` and `from_raw_bytes()` to those binary-wrapper types above which would be identical to their current `to_bytes()`/`from_bytes()`
3. add in `to_cbor_bytes()`and `from_cbor_bytes()` to all types and have them just call `to_bytes()`/`from_bytes()` for most types, but have them include the CBOR binary wrapper around those special hash/key/address types above.

[...]

PR #293, opened 2021-12-13
Adding to_hex/from_hex

alessandrokonrad:
This reduces a little bit of overhead in Js with the usage of Buffer.


rust/src/utils.rs
```diff
@@ -59,11 +91,27 @@ macro_rules! to_bytes {
...
+    }
+}
+
 #[macro_export]
 macro_rules! to_from_bytes {
     ($name:ident) => {
         to_bytes!($name);
         from_bytes!($name);
+        to_hex!($name);
+        from_hex!($name);
```
rooooooooob, line 114: @alessandrokonrad Which format should we be using here for hex? The raw bytes? or the CBOR bytes? Currently `from_bytes()` can return one of those depending on which types. Most types are CBOR bytes, but Addresses, crypto keys and hashes are all raw bytes. See: #263

What is your thoughts on this? I proposed having `to_raw_bytes()` and `to_cbor_bytes()` but then if we add in hex to this too that's a lot of variants.

rooooooooob (commented):
We should figure out #263 before we merge a PR like this.

SebastienGllmt:
Just as a historical note, the reason why we didn't add small utility functions like this is because projects like Yoroi Mobile rely on native bindings to the Rust code. We tried to minimize the amount of time spent on native bindings by trying to have a single way to do things in the Rust side

rust/src/utils.rs
alessandrokonrad: > Which format should we be using here for hex? The raw bytes? or the CBOR bytes?

to_hex/from_hex are simply wrappers around to_bytes/from_bytes. So it returns/takes whatever that is. The macro to_from_bytes is implemented for types that are fully in CBOR I think. I'm not sure how these would look like in raw bytes, for instance `Value`?

> Most types are CBOR bytes, but Addresses, crypto keys and hashes are all raw bytes

I think it makes sense to return them in a format how the cardano ledger would take them, so probably as it is right now. Everything else would be more confusing

rooooooooob: >Would to_cbor_bytes simply add a tag in front of the raw bytes, that says something is bytes?

In the cases of the hashes/keys and addresses, yes. They're all defined in the .cddl as `address = bytes` (usually with size limits) for example. When they're used in the context of the CDDL CBOR encoding (e.g. when composed into the other CBOR structures in that spec) they need to have those extra bytes. This is related to our discussion in cardano-foundation/CIPs#148 which going by how the spec is written would need those bytes, which is something that we might want to change there. Either way it's a bit confusing IMO having these two notions of what a byte representation means and I'm not sure about the best way to go about it, which is why I opened up #263.

>I think it makes sense to return them in a format how the cardano ledger would take them, so probably as it is right now

If we stay with this we should at least document it.

alessandrokonrad: > This is related to our discussion in cardano-foundation/CIPs#148 which going by how the spec is written would need those bytes, which is something that we might want to change there

Yes I would be for a change there. Or we add a simple helper function in either the message signing library or the serialization-lib to add a tag in front of the raw bytes.

lisicky (commented):
/check

lisicky (approved):

vsubhuman:
Pulling this into a local branch so we can work on upgrading it

vsubhuman:
Thank you, @alessandrokonrad !

PR #297, opened 2021-12-16
Simplify minimum UTXO value

SebastienGllmt:
The old code has the following flow:

1. If no asset present, result `34482 * (27 + 2)` = 999,978 (the current min ADA on mainnet for ADA-only txs)
2. If asset present,
  1. If no asset (redundant check), return  `34482 * (27 + 1)` -- note this is incorrect, but nobody noticed since this branch never ran
  2. Calculate the complex minimum value function

The new code deletes (1) and fixes the calculation in (2) to return the right result in the no-ADA case

### DIFF
--- a/rust/src/utils.rs
+++ b/rust/src/utils.rs
@@ -1009,9 +1009,9 @@ fn bundle_size(
     assets: &Value,
     constants: &OutputSizeConstants,
 ) -> usize {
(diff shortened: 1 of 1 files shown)


rooooooooob (approved):
LGTM

vsubhuman (approved):
Thanks, Seba! Looks good

PR #304, opened 2021-12-22
Simplify txbuilder output addition

SebastienGllmt:
Currently a few places in cardano-serialization-lib hardcode that no data hash is present in outputs which can lead to the incorrect minimum ADA amount being calculated.

This PR fixes these by properly tracking data hash in the tx builder.

However, I ran into a problem: the tx builder currently has a combinatorial explosion of different functions for adding outputs. Notably,
- add_output_amount
- add_output_coin
- add_output_asset_and_min_required_coin
- add_mint_asset_and_output
- add_mint_asset_and_output_min_required_coin

We can't just add a new optional parameter to all these functions because WASM doesn't like `Option`, so the naive solution would be to double the number of functions which isn't acceptable either.

To tackle this, I instead created an `TransactionOutputBuilder` so that it's easier to create plain `TransactionOutput` objects based on certain requirements. This simplifies the txbuilder code and also makes it easier to deal with `TransactionOutput` creation on the JS-side as well


SebastienGllmt:
Note: this will soft-conflict with the following PR because it hardcodes `false` in the tx builder
https://github.com/Emurgo/cardano-serialization-lib/pull/290

rooooooooob (approved):
LGTM although @vsubhuman should check it over too since he was the one who added these convenience functions to begin with and his PR #273 touches some of this code (although from what I can see it should be fine since that PR doesn't change how the output is added for the mint+output convenience setters.)

vsubhuman (approved):
Great change, thank you, Seba!

vsubhuman (commented):
/check

PR #306, opened 2021-12-24
Simplify Implicit Input calculation

SebastienGllmt:
This PR just gets rid of duplicate code

### DIFF
--- a/rust/src/utils.rs
+++ b/rust/src/utils.rs
@@ -929,19 +929,7 @@ pub fn internal_get_implicit_input(
                 |acc, ref withdrawal_amt| acc.checked_add(&withdrawal_amt)
             )?,
     };
-    let certificate_refund = match &certs {
-        None => to_bignum(0),
-        Some(certs) => certs.0
-            .iter()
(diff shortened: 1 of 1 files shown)


rust/src/utils.rs
```diff
@@ -929,19 +929,7 @@ pub fn internal_get_implicit_input(
                 |acc, ref withdrawal_amt| acc.checked_add(&withdrawal_amt)
             )?,
     };
-    let certificate_refund = match &certs {
-        None => to_bignum(0),
-        Some(certs) => certs.0
-            .iter()
-            .try_fold(
-                to_bignum(0),
-                |acc, ref cert| match &cert.0 {
```
rooooooooob, line 938: aren't these going in the opposite directions (register vs retirement)? the old code does it for `PoolRetirement`/`StakeDeregistration` but the new code would use `PoolRegistration` and `StakeRegistration`.

SebastienGllmt: Good catch. Somehow I missed this

PR #307, opened 2021-12-25
Add Witness builder

SebastienGllmt:
The problem with `TransactionWitnessSet` is that at the protocol level it should be a set, but at the CDDL level it's just a list. This means it's up to individual implementations to de-duplication which is not ideal.

This PR adds a TransactionWitnessSetBuilder which handles the deduplication for you. It's a first step to #37 and can also help with #273

### DIFF
--- a/rust/src/witness_builder.rs
+++ b/rust/src/witness_builder.rs
@@ -1,0 +1,225 @@
+use std::collections::HashMap;
+use super::*;
+
+/// Builder de-duplicates witnesses as they are added 
+#[wasm_bindgen]
+#[derive(Clone)]
+pub struct TransactionWitnessSetBuilder {
+    // See Alonzo spec section 3.1 which defines the keys for these types
+    vkeys: HashMap<Vkey, Vkeywitness>,
+    bootstraps: HashMap<Vkey, BootstrapWitness>,
+    native_scripts: HashMap<ScriptHash, NativeScript>,
+    plutus_scripts: HashMap<ScriptHash, PlutusScript>,
+    plutus_data: HashMap<DataHash, PlutusData>,
+    redeemers: HashMap<RedeemerTag, Redeemer>,
+}
+
+#[wasm_bindgen]
+impl TransactionWitnessSetBuilder {
+    pub fn add_vkey(&mut self, vkey: &Vkeywitness) {
+        self.vkeys.insert(vkey.vkey(), vkey.clone());
+    }
+
+    pub fn add_bootstrap(&mut self, bootstrap: &BootstrapWitness) {
+        self.bootstraps.insert(bootstrap.vkey(), bootstrap.clone());
+    }
+
+    pub fn add_native_script(&mut self, native_script: &NativeScript) {
+        self.native_scripts.insert(native_script.hash(ScriptHashNamespace::NativeScript), native_script.clone());
+    }
+
+    pub fn add_plutus_script(&mut self, plutus_script: &PlutusScript) {
+        // TODO: don't assume PlutusV1 and instead somehow calculate this
+        self.plutus_scripts.insert(plutus_script.hash(ScriptHashNamespace::PlutusV1), plutus_script.clone());
+    }
+
+    pub fn add_plutus_datum(&mut self, plutus_datum: &PlutusData) {
+        self.plutus_data.insert(hash_plutus_data(&plutus_datum), plutus_datum.clone());
+    }
+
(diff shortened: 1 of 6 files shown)


SebastienGllmt:
I went ahead and also added a `RequiredSignatureSet` so that the tx builder can pass information about which witnesses are required. I didn't hook it up to the tx builder since probably it would conflict with #273 too much

rust/src/witness_builder.rs
```diff
@@ -0,0 +1,393 @@
...
+            RedeemerWitnessKey::new(&redeemer.tag(), &redeemer.index()),
+            redeemer.clone()
+        );
+    }
+
+    pub fn set_required_sigs(&mut self, required_sigs: &RequiredSignatureSet) {
+        self.required_sigs = required_sigs.clone()
+    }
+
+    pub fn get_required_sigs(&self) -> RequiredSignatureSet {
```
SebastienGllmt, line 178: I'm not sure if we want to expose this getter since it's misleading (it returns a copy of a mutable object -- not a reference)

SebastienGllmt: I ended up removing it since I feel it's just a foot-gun

rust/src/witness_builder.rs
```diff
@@ -0,0 +1,389 @@
...
+        );
+        builder.add_native_script(&wit1);
+        builder.add_native_script(&wit1);
+
+        // add a different element
+        builder.add_native_script(&NativeScript::new_timelock_start(
+            &TimelockStart::new(2),
+        ));
+
+        let wit_set = builder.build().unwrap();
```
rooooooooob, line 359: We have the not-enough-errors case tested in `requirement_test()` for failure but not for success anywhere. All of these tests here were written before `RequiredWitnessSet` was added so they don't use it. We should modify at least one of these tests (or all or write a separate one) to include a `RequiredWitnessSet` to test that path too as these will all succeed since there are no required things.

rust/src/witness_builder.rs
```diff
@@ -0,0 +1,389 @@
...
+            index: index.clone(),
+        }
+    }
+
+}
+
+#[wasm_bindgen]
+#[derive(Clone)]
+pub struct RequiredWitnessSet {
+    vkeys: HashSet<Vkey>,
```
rooooooooob, line 34: I am assuming we will be constructing `RequiredWitnessSet` within the `TxBuilder`? If so, how are we supposed to get the `Vkey`? Wouldn't we only have the hash of it in addresses? Or are we gonna need to give it that info somehow like modifying `TxBuilderInput`? Or store the`Ed25519KeyHash` here instead? Although crypto calls are by far the slowest thing in this library so that could be a point of consideration.

SebastienGllmt: Hmmm. The actual key for uniqueness used in the witness data structure is the `VKey` but the mapping between `VKey` <-> `Ed25519KeyHash` should be unique in practice so you're right that we can deviate from the spec to make the interface more usable from the tx builder. I don't think this should cause any problems

rust/src/witness_builder.rs
SebastienGllmt: We have  `requirement_test` for testing the `RequiredWitnessSet` but I added a `requirement_test_pass` that makes sure the success case works for everything

vsubhuman (commented):
/check

rust/src/witness_builder.rs
```diff
@@ -0,0 +1,415 @@
...
+    // See Alonzo spec section 3.1 which defines the keys for these types
+    vkeys: HashMap<Vkey, Vkeywitness>,
+    bootstraps: HashMap<Vkey, BootstrapWitness>,
+    native_scripts: HashMap<ScriptHash, NativeScript>,
+    plutus_scripts: HashMap<ScriptHash, PlutusScript>,
+    plutus_data: HashMap<DataHash, PlutusData>,
+    redeemers: HashMap<RedeemerWitnessKey, Redeemer>,
+
+    // witnesses that need to be added for the build function to succeed
+    required_wits: RequiredWitnessSet,
```
vsubhuman, line 149: @SebastienGllmt , can you plz leave a lil comment, what's the idea with this field, i.e. how and when it is presumed to be used?

nicolasayotte:
This is important. Wen?

jonathangenlambda:
Would it be possible to add functionality to combine two `TransactionWitnessSet`?

PR #314, opened 2022-01-05
Bump wasm-bindgen

SebastienGllmt:
Version  0.2.75 of wasm-bindgen fixes a bug when using nodejs + webpack v5 but we're still on 0.2.74 so I bump the version


SebastienGllmt:
To be clear, this doesn't include the fix for the stack space leak in wasm-bindgen (they haven't pushed a release for this yet)

SebastienGllmt:
We use this branch in multiple places and no issue so far

vsubhuman (approved):
Thank you, @SebastienGllmt !

PR #318, opened 2022-01-08
[9.1.3] MetadataMap deserialization fix

vsubhuman:
Cardano blockchain itself does allow duplicated keys in metadata map as does CBOR itself, but we had an assertion to fail in case a duplicated key been encountered, which causes crashes and inability to parse some existing transactions with their metadata.


rooooooooob:
It looks like only CBOR strict mode disallows duplicates. I wonder where else this would be an issue. For user-generated maps there's also plutus datums but maybe the plutus spec disallows that?

vsubhuman:
New versions published:

- https://www.npmjs.com/package/@emurgo/cardano-serialization-lib-nodejs/v/9.1.3
- https://www.npmjs.com/package/@emurgo/cardano-serialization-lib-browser/v/9.1.3
- https://www.npmjs.com/package/@emurgo/cardano-serialization-lib-asmjs/v/9.1.3
- https://crates.io/crates/cardano-serialization-lib/9.1.3

vsubhuman:
Release: https://github.com/Emurgo/cardano-serialization-lib/releases/tag/9.1.3

PR #319, opened 2022-01-10
Emphasize that this library is written in Rust

rodrigc:
Emphasize a bit more that this is a Rust library.
(For newbies like me who stumble across this library via web search).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Cardano Serialization Lib
 
-This is a library for serialization & deserialization of data structures used in Cardano's Haskell implementation of Shelley along with useful utility functions.
+This is a library, written in Rust, for serialization & deserialization of data structures used in Cardano's Haskell implementation of Shelley along with useful utility functions.
 
 ##### NPM packages
 


gavinharris-dev:
Would this be correct? Would be it more correct to say that it is a "Library written in Rust" as the 'WASM' build can also be utilised from JavaScript/TypeScript within both NodeJS and the Browser.

rodrigc:
Sure, I'm open to rephrasing.  My main intent is to try to boost this library in search engine results for **Cardano Rust**.
I brought this up on the dcSpark Discord: https://discord.com/channels/857911582972968981/881954687211933696/929906586900631592

rodrigc:
I also brought up similar issues about web search for **Cardano Rust** on the IOG Discord: https://discord.com/channels/826816523368005654/894843835023228950/929914460754280528

rooooooooob:
As mentioned above, it's written *using* rust, but the vast, vast majority of people using this library are using the wasm or asm.js bindings for it. For searches, the readme does mention rust a bit below. I guess multiple references could help it though? The readme in general needs to be updated as it still mentions "haskell shelley" (to differentiate itself from the jormungandr shelley testnet in rust) even thuogh we've updated the library to work with Alonzo since then.

rodrigc:
OK, I changed to **written in Rust**.

rodrigc:
I would suggest that:
**used in Cardano's Haskell implementation of Shelley**

to something like:

**used in Cardano's implementation of Alonzo**

SebastienGllmt (approved):

rodrigc:
Thanks for merging this.  Is there a way to boost this library in this list: https://github.com/trending/rust ?

That way automated bots such as https://twitter.com/RustTrending will Tweet about this library.

PR #330, opened 2022-01-14
Deprecating `TransactionBody::multiassets`

vsubhuman:


rust/src/lib.rs
```diff
@@ -305,10 +305,20 @@ impl TransactionBody {
         self.mint = Some(mint.clone())
     }
 
-    pub fn multiassets(&self) -> Option<Mint> {
+    pub fn get_mint(&self) -> Option<Mint> {
```
rooooooooob, line 308: It would fit better in with the rest of the library if it were named just `mint()` (i.e. rust naming conventions).

vsubhuman: Good point, changing this now. Was worried it sounds too much like a verb, e.g. `tx.mint();` but shouldn't be a problem.

vsubhuman (commented):
/check

PR #332, opened 2022-01-15
fix serialization for BigInt 0

longngn:
Basically, `u64_digits` will be `[]` when BigInt is 0, not `[0]`
This lead to wrong serialization of datums such as `{"int":0}`


rooooooooob (approved):
LGTM, thanks for the PR!

rooooooooob:
@vsubhuman  I merged this because we have too many open PRs and it was a simple fix and is closer to what my intention was when I wrote this code originally, I just missed the edge case.

PR #345, opened 2022-01-26
JSON support

rooooooooob:
This also auto-generates .ts (and thus flow) definitions for all types and inserts them into the `to_js_value(): JsValue` .ts/flow defs in place of `any`.

This is a multi-phase process:
1) Generate (using `schemars` traits) .json schemas for all types. This
is done by running the program in rust/json-gen which will export these
to rust/json-gen/schemas
2) Running scripts/run-json2ts.js to convert these to .ts files. This also does a lot of fixing up like appending JSON, and removing weird warts generated by the 2 tools (schemars/json-schemas->ts)
3) Running scripts/json-ts-types.js to replace any `any` return types
with the correct *JSON type generated above and merging these into the main .ts def file


scripts/run-json2ts.js
```diff
@@ -0,0 +1,13 @@
...
+const json2ts = require('json-schema-to-typescript');
+const path = require('path');
+
+const schemasDir = path.join('rust', 'json-gen', 'schemas');
+const schemaFiles = fs.readdirSync(schemasDir).filter(file => path.extname(file) === '.json');
+
+Promise.all(schemaFiles.map(schemaFile => {
+  const completeName = path.join(schemasDir, schemaFile);
+  console.log(`Processing: ${completeName}`);
+  return json2ts.compileFromFile(completeName, {declareExternallyReferenced: false});
```
rooooooooob, line 11: I thought `declareExternallyReferenced` would help but I guess they're inlined so it didn't help

rooooooooob:
More byte types should probably be manually declared as hex strings too since schemars/serde default to a number array.

rust/Cargo.toml
```diff
@@ -53,3 +55,8 @@ quickcheck = "0.9.2"
 quickcheck_macros = "0.9.1"
 rand_os = "0.1"
 rand_chacha = "0.1"
+
+[build-dependencies]
```
rooooooooob, line 59: We can probably get rid of these. I was experimenting with rust build scripts but decided to just have a subcrate to run the script instead since I couldn't access the main crate from it (but I've never worked with rust build scripts so maybe it's possible). It makes more sense to only run these when we do the wasm builds/flowgen I think though anyway.

rust/src/metadata.rs
```diff
@@ -219,6 +219,32 @@ impl TransactionMetadatum {
...
+    D: serde::de::Deserializer<'de> {
+        let s = <String as serde::de::Deserialize>::deserialize(deserializer)?;
+        encode_json_str_to_metadatum(s.clone(), MetadataJsonSchema::DetailedSchema)
+            .map_err(|e| serde::de::Error::invalid_value(serde::de::Unexpected::Str(&s), &format!("{:?}", e).as_str()))
+    }
+}
+
+// just for now we'll do json-in-json until I can figure this out better
+// TODO: maybe not generate this? or how do we do this?
+impl JsonSchema for TransactionMetadatum {
```
rooooooooob, line 242: This is not the best solution but the options are either:
1) Just make our own (incompatible with cardano-node) JSON rep using schemars/serde
2) Manually write .ts defs for the cardano-node JSON schema we follow (I picked `DetailedSchema` since it supports everything (besides edge-cases with `i64` in serde-json) and merge those in as a final step. We'd want to do the same with plutus datums I guess if we go this route.

I'm leaning towards 2 but wanted to get opinions on it. We might have to write them as JSON schemas and generate from there and hope we can get it to match the .ts we want as otherwise other schemas referring to metadata (e.g. tx) might break.

rust/src/utils.rs
```diff
@@ -67,6 +69,30 @@ macro_rules! to_from_bytes {
...
+        #[wasm_bindgen]
+        impl $name {
+            pub fn to_json(&self) -> Result<String, JsError> {
+                serde_json::to_string_pretty(&self)
+                    .map_err(|e| JsError::from_str(&format!("to_json: {}", e)))
+            }
+
+            #[cfg(all(target_arch = "wasm32", not(target_os = "emscripten")))]
+            pub fn to_js_value(&self) -> Result<JsValue, JsError> {
+                JsValue::from_serde(&self)
```
rooooooooob, line 84: See comment above in the cargo.toml

rust/src/utils.rs
```diff
@@ -243,6 +269,27 @@ impl Deserialize for BigNum {
   }
 }
 
+impl serde::Serialize for BigNum {
```
rooooooooob, line 272: Do we want `to_json()`/`to_js_value()`/`from_json()` here (and BigInt, etc) for consistency even though `to_str()`/`from_str()` do the same thing as this would do?

scripts/json-ts-types.js
```diff
@@ -0,0 +1,37 @@
...
+  //const toJson = /\sto_json\(\): any;/.exec(line);
+  //console.log(toJson);
+  inputFile[i] = line.replace(/(\s?to_js_value\(\)\s?:\s?)(any)(;)/, `$1${currentClass}JSON$3`);
+  if (line != inputFile[i]) {
+    continue;
+  }
+  // TODO: we might want to make sure we don't have other cases where this would replace
+  // things it shouldn't. We'd have to do some go-back-a-few-lines replace to only do this
+  // for to_json() comments.
+  inputFile[i] = line.replace(/(\s?\*\s?\@returns\s\{)(any)(\})/, `$1${currentClass}JSON$3`);
```
rooooooooob, line 24: None of the types return `any` in our library for now but this could be a problem in the future.

SebastienGllmt (approved):

lisicky:
@rooooooooob could you update cardano_serialization_lib.js.flow ? I'm afraid I do smth wrong because I got different result then in the branch.

SebastienGllmt:
@lisicky dcSpark will no longer be contributing to this repo because we are instead managing our fork (cardano-multiplatform-lib) and it would be too much work to back-port all our changes back to here.

rooooooooob:
@lisicky it's a generated file you shouldn't need to manually edit it

lisicky (commented):
/check

lisicky (commented):
/check

lisicky (approved):

lisicky (commented):
/check

lisicky (commented):
/check

lisicky (commented):
/check

vsubhuman (approved):

PR #347, opened 2022-02-01
Fix hash_script_data() for no-redeemers special case

rooooooooob:
See input-output-hk/cardano-ledger#2606 for documentation of this
edge-case

and input-output-hk/cardano-ledger#2604 for the original issue


SebastienGllmt (approved):

vsubhuman (approved):
Thank you, @rooooooooob !

PR #354, opened 2022-02-02
bump version of cryptoxide and ed25519-bip32

vincenthz:


vsubhuman (commented):
/check

rust/src/chain_crypto/algorithms/ed25519.rs
```diff
@@ -11,7 +13,7 @@ use ed25519_bip32::XPub;
 pub struct Ed25519;
 
 #[derive(Clone)]
-pub struct Priv([u8; ed25519::SEED_LENGTH]);
+pub struct Priv([u8; ed25519::PRIVATE_KEY_LENGTH]);
```
vsubhuman, line 16: Just want to make sure. I understand this change will make all instance of this `Priv` to be 64 bytes instead of 32, but at the same time I am not sure if this specific `Ed25519` is used for anything at all, directly because I can only see `Ed25519Bip32` being used for all the function that are in actual use.

@rooooooooob , can you plz clarify what the `Ed25519` is being used for , compared to `Ed25519Bip32`, just so that it's extra clear about the security of this change?

vsubhuman: > PRIVATE_KEY_LENGTH is 64 bytes in cryptoxide 0.3, but 32 bytes in cryptoxide 0.4

image

Understood! This is clear now, thank you, @vincenthz !

vsubhuman (approved):

rooooooooob:
Nice, this should help with performance. Hopefully it carries over to the wasm and asm.js sides too.

PR #365, opened 2022-02-10
[10.0.3] Added function `NativeScript.get_required_signers()`

vsubhuman:
The new function returns an array of all **unique** key-hashes used in the script **recursively**. So e.g.

```
All(
  Key(X),
  Any(
    Key(Y),
    TimelockExpiry(42),
  )
)
```

Calling `get_required_signers()` on the top script will return `[X, Y]` (the order is not guaranteed in any way).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "cardano-serialization-lib",
-  "version": "10.0.2",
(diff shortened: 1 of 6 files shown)


vsubhuman (commented):
/check

PR #367, opened 2022-02-10
PointerAddress support for >u32::MAX

### LINKED ISSUE #364: Can't decode address addr_test1grqe6lg9ay8wkcu5k5e38lne63c80h3nq6xxhqfmhewf645pllllllllllll7lupllllllllllll7lupllllllllllll7lc9wayvj
Today we stumble upon this address: `addr_test1grqe6lg9ay8wkcu5k5e38lne63c80h3nq6xxhqfmhewf645pllllllllllll7lupllllllllllll7lupllllllllllll7lc9wayvj`
This is a valid address on-chain: https://testnet.cardanoscan.io/address/40c19d7d05e90eeb6394b53313fe79d47077de33068c6b813bbe5c9d5681ffffffffffffffff7f81ffffffffffffffff7f81ffffffffffffffff7f
However, when we use `Address.from_bech32`, we get this error:
```
(node:252137) UnhandledPromiseRejectionWarning: Deserialization failed in Address.Address.Pointer.slot because: Invalid cbor: expected 32bit long unsigned integer
(Use `node --trace-warnings ...` to show where the warning was created)
```

rooooooooob:
Fixes #364 in a way that requires the least changes.

Only users who are manually creating `Pointer` to make a new
`PointerAddress` from parts will be affected by this API change.

We could also alternatively change *ALL* references of `Slot` and `TransactionIndex`
to support `BigNum` instead of `u32`. This would be inline with the
alonzo.cddl binary spec, but would make use of the library more
annoying, especially from javascript/wasm, as well as requiring more code changes.

I don't see why you would have a transaction index greater than even a
single byte, let alone 2^32 - 1. But since that's how the binary spec
is, it's possible we'll run into on-chain data that do this in the future.

This isn't the only place where we run into issues like this  as there are
edge cases anywhere where `nint` is specified since the CBOR library we
support doesn't handle numbers between `i64::MIN` and `-u64::MAX`.


vsubhuman (approved):
Thank you, @rooooooooob !

Issue #371, opened 2022-02-14
struction of Costmdls is wrong? 


somereason:
Hi

I'm trying to redeem asset from contract utxo. I was success via cardano-client-lib(java), and almost success via cardano-serialization-lib(javascript)

The issue holding me is the calculation of hash_script_data. I compared java version and js version, foud the differance on Costmdls.

on java version. it is hard coded in package com.bloxbean.cardano.client.transaction.util.CostModelConstants.LANGUAGE_VIEWS, value is 
```
a141005901d59f1a000302590001011a00060bc719026d00011a000249f01903e800011a000249f018201a0025cea81971f70419744d186419744d186419744d186419744d186419744d186419744d18641864186419744d18641a000249f018201a000249f018201a000249f018201a000249f01903e800011a000249f018201a000249f01903e800081a000242201a00067e2318760001011a000249f01903e800081a000249f01a0001b79818f7011a000249f0192710011a0002155e19052e011903e81a000249f01903e8011a000249f018201a000249f018201a000249f0182001011a000249f0011a000249f0041a000194af18f8011a000194af18f8011a0002377c190556011a0002bdea1901f1011a000249f018201a000249f018201a000249f018201a000249f018201a000249f018201a000249f018201a000242201a00067e23187600010119f04c192bd200011a000249f018201a000242201a00067e2318760001011a000242201a00067e2318760001011a0025cea81971f704001a000141bb041a000249f019138800011a000249f018201a000302590001011a000249f018201a000249f018201a000249f018201a000249f018201a000249f018201a000249f018201a000249f018201a00330da70101ff
```

when you decode it with cbor, you can find it has the struct of 
```
{
  bytes, //language version
  cbor<unsigned[]> //value of (testnet-alonzo-genesis.json costModels.PlutusV1), sorted by key
}
```

![图片](https://user-images.githubusercontent.com/3340724/153870660-12e36114-2902-4972-8224-0b4c43a3a37f.png)

here is result of decoding inner cbor
![图片](https://user-images.githubusercontent.com/3340724/153871062-67496a5a-3ab8-4443-83b8-6fdbd8f49820.png)

 But when I check the bytes generated by js version, the binary is 
```
[...]

PR #373, opened 2022-02-14
Fixing hash_script_data

vsubhuman:
The code is updated to replicate the hashing from this test case: https://github.com/bloxbean/cardano-client-lib/blob/1bb66ca7fba785d3f44909d7472639df575a9163/src/test/java/com/bloxbean/cardano/client/transaction/util/ScriptDataHashGeneratorTest.java

Which is reported by users as known to be correct: https://github.com/Emurgo/cardano-serialization-lib/issues/371


rust/src/plutus.rs
```diff
@@ -475,6 +475,13 @@ impl PlutusList {
         }
     }
 
+    pub(crate) fn clone_definite(&self) -> Self {
+        Self {
+            elems: self.elems.clone(),
+            definite_encoding: Some(true),
+        }
+    }
```
vsubhuman, line 483: @rooooooooob , I am not 100% sure why indefinite serialization been implemented for the plutus list in the first place, and whether it is correct to be used in any other case apart from the `hash_script_data` function. It might be that it SHOULD be indefinite everywhere else, but because of an error in the hashing - it has to be definite in `hash_script_data`. Or it could be that it actually should be definite everywhere at all, and then maybe the `Serialize` implementation for the `PlutusList` must be changed to always do definite serializing. 

So far it seems that plutus data hashes are correct between the serlib and CLI, we'll be rechecking it for `PlutusList` now

vsubhuman:
Received two confirmations so far that this fixed version works for script inputs now: https://github.com/Emurgo/cardano-serialization-lib/issues/371#issuecomment-1039344520 and https://github.com/Emurgo/cardano-serialization-lib/issues/371#issuecomment-1039345094

rooooooooob (commented):
@vsubhuman I'm not sure this was actually an error, or if this fix is correct. Plutus datums are NOT converted to canonical CBOR form according to IOHK (sadly - I wish they did since this has caused so many headaches....) when hashed, but instead hashed as-is. I'm not 100% sure if this applies to this `ScriptDataHash`, or just `DatumHash` though.

The comment above (taken from alonzo.cddl originally) says
```
; script data format:
; [ redeemers | datums | language views ]
; The redeemers are exactly the data present in the transaction witness set.
; Similarly for the datums, if present. If no datums are provided, the middle
; field is an empty string.
```

Which sounds like they should be the same as in the witness set. This, however, would potentially make that untrue. It's also possible that this only applies to *each datum itself* and not the datum list itself, but it would be good to find out.

However it seems that @somereason was using the same list in both the witness set and and passing to this.

What was the difference in 10.0.4-beta-1? (which seemed to fix their issue?) I only see beta-2 referenced in this PR.

vsubhuman:
> What was the difference in 10.0.4-beta-1? (which seemed to fix their issue?) I only see beta-2 referenced in this PR.

Beta.2 is unrelated and only available on Crates, sorry, beta.1 actually includes the fix and is the latest one on NPM atm. So the users are reporting `beta.1` does work and they been able to submit the txs accepted by the nodes.

> The comment above (taken from alonzo.cddl originally) says
>
>> ; script data format:
>> ; [ redeemers | datums | language views ]
>> ; The redeemers are exactly the data present in the transaction witness set.
>> ; Similarly for the datums, if present. If no datums are provided, the middle
>> ; field is an empty string.
> 
> Which sounds like they should be the same as in the witness set. This, however, would potentially make that untrue. It's also possible that this only applies to each datum itself and not the datum list itself, but it would be good to find out.

The thing is that when the witnes-set is serialized - the field 4 is always hardcoded to serialized as definite: https://github.com/Emurgo/cardano-serialization-lib/blob/1eb7fa90be3c3f436136737271ff65ee14efdc71/rust/src/serialization.rs#L1808-L1812

So there's no check happening whether the list is definite or not. So the serialization as always definite in the `hash_script_data` after this change matches what we are doing for the field 4.

vsubhuman:
> It's also possible that this only applies to _each datum itself_ and not the datum list itself, but it would be good to find out.

@rooooooooob , this is the main reason why I have hesitated to get this merged and released without some further testing and investigation. So the `PlutusList` type is used at the top-level holder for the datums in top-level places, like the field-4 in the witness set and when passing it to the `hash_script_data` - and in both cases it is correct to serialize it as definite.

The remaining question is that the `PlutusList` is also a type for a separate `PlutusData`, which might be for example a part of a higher level datum, e.g. a value in a map, or an entry in another list. And I want to run some testing to double-check if indefinite serialization makes sense in this case.

If yes - then some clarifications might need to be added to the CDDL. But if not - I think we oughtta change the `PlutusList` serialization implementation then to always be definite.

rust/src/utils.rs
```diff
@@ -921,7 +921,7 @@ pub fn hash_script_data(redeemers: &Redeemers, cost_models: &Costmdls, datums: O
         */
         buf.extend(redeemers.to_bytes());
         if let Some(d) = &datums {
-            buf.extend(d.to_bytes());
+            buf.extend(d.clone_definite().to_bytes());
```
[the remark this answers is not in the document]
vsubhuman: The answer to this is - yes, after looking into the details described in this comment: https://github.com/Emurgo/cardano-serialization-lib/pull/373#issuecomment-1039498195

The serialization in both branches of the `hash_script_data` is expected to match the witness-set field-4 which means the list must be serialized as definite.

vsubhuman:
> The remaining question is that the `PlutusList` is also a type for a separate `PlutusData`, which might be for example a part of a higher level datum, e.g. a value in a map, or an entry in another list. And I want to run some testing to double-check if indefinite serialization makes sense in this case.
> 
> If yes - then some clarifications might need to be added to the CDDL. But if not - I think we oughta change the `PlutusList` serialization implementation then to always be definite.

@rooooooooob , I am guessing by the existence of https://github.com/Emurgo/cardano-serialization-lib/pull/228 that the correct answer is that yes - when PlutusList is a part of datum - it must be indefinite, but when it's field-4 - it must be definite, and by extension it also must be definite when passed to `hash_script_data`.

So it seems the fix in its current form must be the correct version, logically. Maybe the code can be better.

rust/src/plutus.rs
vsubhuman: Seems to be resolved by this: https://github.com/Emurgo/cardano-serialization-lib/pull/373#issuecomment-1039513926

vsubhuman (commented):
/check

rooooooooob:
@vsubhuman I asked IOHK about the format and I'm just waiting on a response. The wording in the script data format is a little ambiguous so it would be great to have an exact answer. I'm also unsure if it's **just** the outer list, or if the list happened to be the only thing that had a mismatch (could be likely if users' datums are simple). Hopefully we'll have an answer soon.

I wish they had just made canonical CBOR a requirement for everything since datum hashes not abiding by it has been causing issues for so many people who are using a mix of tools. But then I guess you could say that if users put up a UTXO with a datum hash that wasn't canonical, at least with the current system you'd be able to provide exactly the output to unlock it. Although on that line of thought people can stick whatever hash there and end up perma-locking their UTXO as-is, albeit a lot harder to do accidentally if you don't require canonical CBOR.

rooooooooob:
>The thing is that when the witnes-set is serialized - the field 4 is always hardcoded to serialized as definite: 

Oh, I didn't know that. I thought we would've been delegating to `PlutusList::serialize()`. That could change things, and maybe that's where the fix should be instead if the only requirement is that it must match directly, especially given that users could be setting that from their own deserialized data rather than constructing from the library. Still, it's better to figure out **exactly** what the format is without a doubt.

rooooooooob:
@vsubhuman I got a response from IOHK. It is *exactly* the encoded CBOR that would be in the witness set, and collections of datums keep their encoding just like individual datums. I still think it's better to make the change inside the witness set serialization instead of here though just in case the entire datum set gets used somewhere else in the future.

vsubhuman:
> got a response from IOHK. It is exactly the encoded CBOR that would be in the witness set, and collections of datums keep their encoding just like individual datums. I still think it's better to make the change inside the witness set serialization instead of here though just in case the entire datum set gets used somewhere else in the future.

@rooooooooob Do you mean to make the change to delegate the serialisation of the witness field-4 to the PlutusList:: serialize to also make it indefinite? 

If so, than what is the reason for the #228 change to exist?

vsubhuman:
> > got a response from IOHK. It is exactly the encoded CBOR that would be in the witness set, and collections of datums keep their encoding just like individual datums. I still think it's better to make the change inside the witness set serialization instead of here though just in case the entire datum set gets used somewhere else in the future.
> 
> @rooooooooob Do you mean to make the change to delegate the serialisation of the witness field-4 to the PlutusList:: serialize to also make it indefinite?
> 
> If so, than what is the reason for the #228 change to exist?

@rooooooooob  If I am understanding the emerging conclusion correctly, does it mean IOHK nodes by themselves don't care necessarily how you serialize a `PlutusList` either on the witness-set top level or within a datum, definite length or indefinite, as long as they can take the exact serialized cbor bytes value from the witness set field-4, and get the matching script hash? Is that what you have received from IOHK?

If so, this would mean we pretty much have to decide somehow, which of the serialization methods to use then, and why exactly. And that would also mean different libraries can produce different hashes for the same underlying data and still both be correct from the chain perspective, which is my biggest concern with all this, e.g. when users will attempt to do some library interoperability or to compare results of building the same transaction from different libraries, etc.

@rooooooooob, do you know if there is anything in the CDDL spec that would suggest that either of the methods (definite or indefinite) is preferrable or more "correct" to be used? 🤔 If the individual datums themselves are correct to be serialized as indefinite - then I agree it makes sense to make top-level lists also always indefinite.

This would mean the #228 change was only incorrect to leave the witness-set field-4 serialization as definite, which caused the whole issue with the mismatched hashes after that.

vsubhuman:
Trying to clarify this with IOHK now as well. From my understanding having to serialize the `PlutusList` as INDEFINITE when it's a part of a datum is a necessity caused by IOHK bugs and they plan to make everything definite and canonical in next versions. So if we can make the top-level serialization as DEFINITE right now - that might be a better preparation for the future

vsubhuman:
Yeah, got the same confirmation from IOHK that the idea is correct - different libraries can decide to use whatever serialization approach they want, as long as the hash matches the field-4, and the fact that different libraries thus might produce different correct transaction for the same exact set of inputs is not considered a problem. No preference is set for either definite or indefinite serialization so I will change the PR to rather update the witness-set serializing to match the script hash

rooooooooob (approved):
LGTM

PR #374, opened 2022-02-14
[10.0.5] Treat burn as part of output in tx-builder

vsubhuman:
Added function `tx_builder.get_total_output` and changed the burn value to be added to the outputs instead of being subtracted from the inputs.

The main point for this is to allow users to specify the required burn and then use coin-selection which will automatically include inputs with the assets required for the burning. Not a major case but would make sense for consistency, e.g. I'd expect for this to work if I were to use it.

From the change calculation perspective, nothing should be different as before it would just subtract the burn from the inputs and now it will add it to the outputs, with the same total outcome.


vsubhuman (commented):
/check

vsubhuman (commented):
/check

vsubhuman (commented):
/check

PR #392, opened 2022-03-02
Fix: nint deserialization below i64::MIN

rooooooooob:
cbor_event's `Deserializer::negative_integer()` panics when it
encounters a number below i64::MIN despite numbers down to -u64::MAX - 1
being valid for a CBOR nint. This fix covers both `BigInt` as well as
`Int`.

### DIFF
--- a/rust/src/utils.rs
+++ b/rust/src/utils.rs
@@ -582,6 +582,8 @@ impl Deserialize for Int {
     }
 }
 
+/// TODO: this function can be removed in case `cbor_event` library ever gets a fix on their side
+/// See https://github.com/Emurgo/cardano-serialization-lib/pull/392
 fn read_nint<R: BufRead + Seek>(raw: &mut Deserializer<R>) -> Result <i128, DeserializeError> {
     let found = raw.cbor_type()?;
     if found != cbor_event::Type::NegativeInteger {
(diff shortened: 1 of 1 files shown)


rust/src/utils.rs
```diff
@@ -573,13 +575,28 @@ impl Deserialize for Int {
...
                 cbor_event::Type::UnsignedInteger => Ok(Self(raw.unsigned_integer()? as i128)),
-                cbor_event::Type::NegativeInteger => Ok(Self(raw.negative_integer()? as i128)),
+                cbor_event::Type::NegativeInteger => Ok(Self(read_nint(raw)?)),
                 _ => Err(DeserializeFailure::NoVariantMatched.into()),
             }
         })().map_err(|e| e.annotate("Int"))
     }
 }
 
+fn read_nint<R: BufRead + Seek>(raw: &mut Deserializer<R>) -> Result <i128, DeserializeError> {
```
rooooooooob, line 585: If I can get a fix merged into `cbor_event` we can remove this in the future and just use `Deserializer::negative_integer()` but this should suffice for now.

NicolasDP: arf we should do something in `cbor_event`.

vsubhuman (approved):
Thank you, @rooooooooob !

Issue #400, opened 2022-03-10
`add_change_if_needed` does not calculate fee for NativeScript?


siegfried:
`add_change_if_needed` does not calculate the correct fee for transactions sent from multi-sig addresses. Here is an example:

https://cardano-multi-sig-app-testnet.vercel.app/transactions/hKMAgYJYIG6zIwOrIEHX%2Fh05BTO4Nbiul6Mi0acMA7%2FDtzTyGlz6AAGCglg5AOQS%2Buq87qmsRuz8yOT2uBUUMfIJdTZXihGvyK26Uct0D1jECr5YEqIm8eWO56Vsliz7m98Lf%2BFmGgBMS0CCWDkwAzhDIFjpiFvESzSX6Dy%2Fj975rmF1Wi9NpzNBbgM4QyBY6YhbxEs0l%2Bg8v4%2Fe%2Ba5hdVovTaczQW4aAqwmfwIaAAJ%2BwaEBgYIBgoIAWBx%2BTl8kD6qxHrI2g6e3OlNMVjjSxmvlNL4d2k2lggBYHOQS%2Buq87qmsRuz8yOT2uBUUMfIJdTZXihGvyK319g%3D%3D

Here is the error message from the wallet after I got the two signed and clicked `Submit Transaction`.

```
{code: 2, info: 'Wallet could not send the tx.', message: '"transaction submit error ShelleyTxValidationError…FeeTooSmallUTxO (Coin 175313) (Coin 163521))))])"'}
```

Here is the source code of building transaction:

https://github.com/siegfried/cardano-multi-sig-app/blob/14058e0fd1982f074932bf5d83392f96bc289d88/components/transaction.tsx#L275

I assume the reason is that the native script is also required by the calculation? How am I supposed to set that in `TransactionBuilder`?

siegfried:
I think the way to [add native script](https://github.com/siegfried/cardano-serialization-lib/blob/c0a8f9d232bfe97c845d94a00d10419ae422f67c/rust/src/tx_builder.rs#L318) to the builder can be improved. [To count the right verification key number](https://github.com/siegfried/cardano-serialization-lib/blob/c0a8f9d232bfe97c845d94a00d10419ae422f67c/rust/src/tx_builder.rs#L64), the native script is required by the `TransactionBuilder`, but the only current available way to include a native script is to mint some assets. The way to add native script can be more general so we can spend coins in a native script.

siegfried:
[...]

PR #402, opened 2022-03-11
Add set_native_scripts() to TransactionBuilder

siegfried:
This PR is to fix issue #400. `NativeScripts` is not only need by minting but also spending on Multi-Sig scripts.

### DIFF
--- a/rust/src/tx_builder.rs
+++ b/rust/src/tx_builder.rs
@@ -61,7 +61,7 @@ fn fake_raw_key_public() -> PublicKey {
 
 fn count_needed_vkeys(tx_builder: &TransactionBuilder) -> usize {
     let input_hashes = &tx_builder.input_types.vkeys;
-    match &tx_builder.mint_scripts {
+    match &tx_builder.native_scripts {
         None => input_hashes.len(),
         Some(scripts) => {
             // Union all input keys with minting keys
@@ -116,7 +116,7 @@ fn fake_full_tx(tx_builder: &TransactionBuilder, body: TransactionBody) -> Resul
             Some(result)
         },
     };
-    let full_script_keys = match &tx_builder.mint_scripts {
+    let full_script_keys = match &tx_builder.native_scripts {
         None => script_keys,
         Some(witness_scripts) => {
             let mut ns = script_keys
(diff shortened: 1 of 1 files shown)


siegfried:
Also, I don't think you can fake a `NativeScripts` with the same size by mere `ScriptHash` [here](https://github.com/Emurgo/cardano-serialization-lib/blob/c0a8f9d232bfe97c845d94a00d10419ae422f67c/rust/src/tx_builder.rs#L99). However, when spending UTxOs from a `NativeScript`, the `NativeScript` is required to be included in the transaction, right? With `set_native_scripts()` added, the size should be able to be calculated [here](https://github.com/Emurgo/cardano-serialization-lib/blob/c0a8f9d232bfe97c845d94a00d10419ae422f67c/rust/src/tx_builder.rs#L119). We could just validate the presence of the `NativeScript` in `add_script_input()`.

siegfried:
In case there might be an update which enables Cardano nodes caching scripts. I suggest to delete [this part](https://github.com/siegfried/cardano-serialization-lib/blob/44469b30e40d52587767cfbe6c0ffd5eaa095628/rust/src/tx_builder.rs#L96) and push the validation to the node.

siegfried:
How about providing a way to let the user decide how many [VKeys](https://github.com/Emurgo/cardano-serialization-lib/blob/c0a8f9d232bfe97c845d94a00d10419ae422f67c/rust/src/tx_builder.rs#L82) should be used? You might be able to count the keys for native scripts but what about the Plutus scripts? Also for native scripts like `Any` and `KofN` the keys included in the transactions can be reduced to the minimum in order to reduce the fee and network loads.

rust/src/tx_builder.rs
```diff
@@ -801,14 +806,14 @@ impl TransactionBuilder {
 
     /// Returns a copy of the current mint witness scripts in the builder
     pub fn get_mint_scripts(&self) -> Option<NativeScripts> {
```
siegfried, line 808: I'm not intended to break the existing interface but this could be `get_native_scripts` as well.

siegfried:
FYI. We've manage to calculate the correct fee by this PR. We've done tests locally.

rust/src/tx_builder.rs
```diff
@@ -93,14 +93,6 @@ fn fake_full_tx(tx_builder: &TransactionBuilder, body: TransactionBody) -> Resul
             Some(result)
         },
     };
-    let script_keys: Option<NativeScripts> = match tx_builder.input_types.scripts.len() {
-        0 => None,
-        _x => {
-            // TODO: figure out how to populate fake witnesses for these
```
siegfried, line 99: You can't.

vsubhuman (commented):
/check

vsubhuman:
Thank you for the suggested change, @siegfried ! But this PR will be closed in favour of the new PR: https://github.com/Emurgo/cardano-serialization-lib/pull/436 , which is supposed to add native scripts handling in a consistent way along with plutus scripts handling

PR #404, opened 2022-03-16
Revert "change undefined to null in tx example"

SebastienGllmt:
This reverts commit cff614b4bcb886e9bf33790c3a81ca71f8b0a7fb

For some reason, it changed the parameter of `Transaction.new` in the example to use `null` instead of `undefined`. This is incorrect and will give you a typescript type error if you try and use null. Not sure why this change was made to the docs.


vsubhuman (approved):
Thanks, Seba!

PR #405, opened 2022-03-16
TLL is too Small

### LINKED ISSUE #394: Slot defined as u32 but tx TTL can be u64 on-chain, resulting in deserialisation errors
While monitoring incoming blocks I have recently encountered some transactions with TTLs larger than `u32`, which results in a deserialisation error when calling `Transaction::from_bytes(..)`. The CDDL specifies slot as `uint`. Is this a known issue?

rvcas:
closes #394 

cc @rooooooooob


rust/src/lib.rs
```diff
@@ -95,7 +95,7 @@ impl UnitInterval {
 type SubCoin = UnitInterval;
 type Rational = UnitInterval;
 type Epoch = u32;
-type Slot = u32;
+type Slot = u64;
```
rooooooooob, line 98: For use from wasm it would be better if we could use `BigNum` instead of `u64`, as `u64` only bindgens to JS `BigInt` in some environments, but doesn't work from others as js numbers only support ints up to I think 53 bits (IEEE doubles).

rooooooooob (changes requested):

rooooooooob (changes requested):
I think you will need to take in every `Slot` by reference now to avoid weird memory problems from wasm_bindgen. I know `BigNum` is `Copy` but I vaguely remember from previous testing that despite this it still moves the object out of the wasm linear memory leading to an invalid pointer if any references from JS remain from it.

rvcas:
ok no problem @rooooooooob

PR #410, opened 2022-03-25
Deserialization fixe for Update + support for Pre-Alonzo Blocks

rooooooooob:
Update's ProtocolVersion was generated as an array due to cddl-codegen
not supporting occurencies so it treated [protocol_version] as an array
rather than a single one as all other occurences of an array with a
single element were * occurences in the cddl.

As of Alozno, Blocks have a mandatory invalid_transactions field which
makes it one of the 2 non-backwards-compatible (the other being
transaction which we already support deserializing both) era changes to
the CDDL. This includes a fix for this by ignoring it if it's not
present.

### DIFF
--- a/rust/pkg/cardano_serialization_lib.js.flow
+++ b/rust/pkg/cardano_serialization_lib.js.flow
@@ -3744,14 +3744,14 @@ declare export class ProtocolParamUpdate {
   extra_entropy(): Nonce | void;
 
   /**
-   * @param {ProtocolVersions} protocol_version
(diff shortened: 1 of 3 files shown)


vsubhuman:
This is a good change, but it breaks the API, so really we'd have to release it as 11.0, need to plan how to handle

vsubhuman (approved):
Thank you, @rooooooooob !

vsubhuman (commented):
/check

vsubhuman (commented):
/check

PR #436, opened 2022-05-17
Script inputs support

vsubhuman:
New APIs added to the tx_builder:
1. `.add_native_script_input` and `.add_plutus_script_input` for when you are adding inputs directly so can add the required witnesses right away
2. New type `PlutusWitness` is introduced, which encapsulates a plutus scripts, a datum, and a redeemer, required to spend a specific input
3. In case generic `.add_input` or `.add_inputs_from` been used for script inputs - the witnesses for these inputs will be missing and this can be checked with `.count_missing_input_scripts` - this will return the total number of missing script witnesses - **NOTE** - either native or plutus, because just from a script hash it is impossible to find out which script kind is it.
4. In order to add the missing script witnesses - `.add_required_native_input_scripts` and `add_required_plutus_input_scripts` can be used, they will return the new count of remaining missing witnesses
5. To acquire already present script witnesses - `.get_native_input_scripts` and `.get_plutus_input_scripts` can be used
6. To handle the plutus data hash value - `.calc_script_data_hash` can be used to calculate the hash automatically from the passed cost model, or alternatively `.set_script_data_hash` and `.remove_script_data_hash` can be used to handle the value manually
7. The existing `.build_tx` is updated to now raise an error in case there are any missing script witnesses in the builder. Alternatively, new `.build_tx_unsafe` can be used to force the transaction building.
8. `PlutusScript.hash()` is now available as well, to calculate a `ScriptHash` from a plutus script, the important part is to know how to deserialise the raw script correctly.

### `NativeScript.hash()` now does not take any arguments!

Previously it was accepting a `ScriptHashNamespace` enum value, and there was only one possible value. Now the enum is extended with another value, but it makes no sense for the `NativeScript` type to optionally accept either of the types - correct hash is only calculated with one namespace value for native scripts and with another namespace value for plutus scripts - so now the namespace is encapsulated within the hash method itself.

Note, this is not a breaking change for JS, which is why it is decided to add it. It will break for Rust users, but is caught at compile time. It is better than having a confusing function with multiple enum values where only one is correct.


vsubhuman (commented):
/check

vsubhuman:
Finishing tests and clarifying about the Plutus inputs order complications

rust/pkg/cardano_serialization_lib.js.flow
```diff
@@ -2773,10 +2774,9 @@ declare export class NativeScript {
   static from_bytes(bytes: Uint8Array): NativeScript;
 
   /**
-   * @param {number} namespace
    * @returns {ScriptHash}
    */
-  hash(namespace: number): ScriptHash;
```
lisicky, line 2779: Can it broke backward compatibility ?

rust/src/tx_builder.rs
```diff
@@ -305,7 +361,7 @@ impl TransactionBuilderConfigBuilder {
 #[derive(Clone, Debug)]
 pub struct TransactionBuilder {
     config: TransactionBuilderConfig,
-    inputs: Vec<TxBuilderInput>,
+    inputs: Vec<(TxBuilderInput, Option<ScriptHash>)>,
```
lisicky, line 364: Where do we use ScriptHash in that vector ?

rust/src/lib.rs
```diff
@@ -1771,14 +1771,14 @@ to_from_bytes!(NativeScript);
...
+    PlutusScript,
 }
 
 #[wasm_bindgen]
 impl NativeScript {
-    pub fn hash(&self, namespace: ScriptHashNamespace) -> ScriptHash {
+    pub fn hash(&self) -> ScriptHash {
         let mut bytes = Vec::with_capacity(self.to_bytes().len() + 1);
-        bytes.extend_from_slice(&vec![namespace as u8]);
+        bytes.extend_from_slice(&vec![ScriptHashNamespace::NativeScript as u8]);
```
lisicky, line 1781: If we don't define numbers for enum ScriptHashNamespace we can find out problems when order of definitions in ScriptHashNamespace will changed

lisicky (changes requested):

rust/src/tx_builder.rs
```diff
@@ -4247,5 +4461,188 @@ mod tests {
         assert_eq!(ma2_output.get(&policy_id2).unwrap().get(&name).unwrap(), to_bignum(40));
     }
 
+    fn create_base_address_from_script_hash(sh: &ScriptHash) -> Address {
```
lisicky, line 4464: I know name is long, but it's better add 'test' or 'testnet' to name for separate from main logic

rust/src/lib.rs
```diff
@@ -1771,14 +1771,14 @@ to_from_bytes!(NativeScript);
 #[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd)]
 pub enum ScriptHashNamespace {
```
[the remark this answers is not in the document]
vsubhuman: This is used on the client side, including Yoroi itself. I am double-checking atm that it will not cause runtime issues in JS and will decide then.

rust/src/lib.rs
vsubhuman: good point

vsubhuman:
Beta published:
- https://www.npmjs.com/package/@emurgo/cardano-serialization-lib-nodejs/v/10.1.0-beta.1
- https://www.npmjs.com/package/@emurgo/cardano-serialization-lib-browser/v/10.1.0-beta.1
- https://www.npmjs.com/package/@emurgo/cardano-serialization-lib-asmjs/v/10.1.0-beta.1

rust/src/lib.rs
vsubhuman: See new commit: https://github.com/Emurgo/cardano-serialization-lib/pull/436/commits/aa94d0ca9a4e9932b5b272555ac4fe0184eede65

rust/src/lib.rs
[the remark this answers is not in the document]
vsubhuman: @lisicky , so we are not removing this from bindgen for now, for the reason that it could cause runtime fails in JS as the field would become undefined. The changed `NativeScript.hash()` that now does not require the namespace argument does NOT cause runtime issues in old JS code when that namespace is being passed. The namespace can be removed from public access later in some major version

rust/src/lib.rs
```diff
@@ -1774,11 +1774,23 @@ pub enum ScriptHashNamespace {
     PlutusScript,
 }
 
+impl ScriptHashNamespace {
+    pub fn scrip_hash_prefix(&self) -> u8 {
+        match self {
+            ScriptHashNamespace::NativeScript => 0 as u8,
```
lisicky, line 1780: Do we have potential issue if we define number directly in enum ?
For example: 
pub enum ScriptHashNamespace
{
    NativeScript = 0
}

lisicky (commented):
/check

rust/src/lib.rs
vsubhuman: Yup, good point, updated here: https://github.com/Emurgo/cardano-serialization-lib/pull/436/commits/789749c1de0154c51304b64233a7e68d9f1ea957

vsubhuman (commented):
/check

lisicky (approved):

vsubhuman (commented):
/check

PR #450, opened 2022-06-08
[babbage] Script ref, inlined datum for TransactionOutput, new min ada

lisicky:


lisicky (commented):
/check

rust/pkg/cardano_serialization_lib.js.flow
```diff
@@ -2529,6 +2529,57 @@ declare export class MetadataMap {
...
+  /**
+   * @param {ScriptRef} script_ref
+   */
+  set_script_ref(script_ref: ScriptRef): void;
+
+  /**
+   * @returns {BigNum}
+   */
+  calculate_ada(): BigNum;
+}
```
vsubhuman, line 2582: @lisicky , not sure if there's a good reason to have a separate structure and type for the calculator, it has separate setters to accept fields like datum, address, value, ref-script, etc. , but there's already exactly a type for all these fields together - `TransactionOutput`. Think we can just have a function with the `TransactionOutput` as an argument and that's it, otherwise with separate setters it becomes less obvious that you need ALL these fields, to calculate the required value correctly

lisicky: @vsubhuman I agree with you. But main reason for creation the structure was that we have different ways for min ada calculation such as min_pure_ada and min_ada_required and I think it's better to have a one point of min ada calculation for a TransactionOutput to reduce numbers of that functions. We can left the structure for internal usage and create only one function with only TransactionOutput and cost of words/bytes. But if you think that the structure is redundant and useless we can discuss in the slack, because there are some discussable moment in the change estimation function.

rust/pkg/cardano_serialization_lib.js.flow
```diff
@@ -4697,6 +4748,34 @@ declare export class ScriptPubkey {
...
+   * @returns {ScriptRef}
+   */
+  static new_native_script(native_script: NativeScript): ScriptRef;
+
+  /**
+   * @param {PlutusScript} plutus_script
+   * @returns {ScriptRef}
+   */
+  static new_plutus_script(plutus_script: PlutusScript): ScriptRef;
+}
```
vsubhuman, line 4778: Looks good with two constructor functions! 

Please add functions:
- `.is_native_script()`
- `.is_plutus_script()`
- `.native_script()`
- `.plutus_script()`

With getters returning an option, same as in the outputs

rust/src/lib.rs
```diff
@@ -1780,6 +1811,38 @@ enum NativeScriptEnum {
     TimelockExpiry(TimelockExpiry),
 }
 
+#[derive(Debug, Clone, Eq, Ord, PartialEq, PartialOrd)]
+pub enum ScriptRefEnum {
+    NativeScript(NativeScript),
+    PlutusScriptV1(PlutusScript),
+    PlutusScriptV2(PlutusScript),
+}
```
vsubhuman, line 1819: Let's remove `PlutusScriptV2` and leave just one `PlutusScript`. Now that we have discussed that the script version will be within the `PlutusScript` type itself, no reason to have separate enums for them

rust/src/lib.rs
```diff
@@ -1780,6 +1811,38 @@ enum NativeScriptEnum {
...
+
+#[wasm_bindgen]
+impl ScriptRef {
+
+    pub fn new_native_script(native_script: &NativeScript) -> Self {
+        Self(ScriptRefEnum::NativeScript(native_script.clone()))
+    }
+
+    pub fn new_plutus_script(plutus_script: &PlutusScript) -> Self {
+        //TODO add choose right plutus script after updating type
```
vsubhuman, line 1835: After https://github.com/Emurgo/cardano-serialization-lib/pull/450/files#r895454845

Update this to just store the `PlutusScript` in the enum, the script version will be solved on the script level, the ref should be independent of that

rust/pkg/cardano_serialization_lib.js.flow
lisicky: Okay : )

rust/src/serialization.rs
```diff
@@ -556,11 +556,43 @@ impl DeserializeEmbeddedGroup for TransactionInput {
...
 impl cbor_event::se::Serialize for TransactionOutput {
     fn serialize<'se, W: Write>(&self, serializer: &'se mut Serializer<W>) -> cbor_event::Result<&'se mut Serializer<W>> {
-        serializer.write_array(cbor_event::Len::Len(if self.data_hash.is_some() { 3 } else { 2 }))?;
-        self.address.serialize(serializer)?;
-        self.amount.serialize(serializer)?;
-        if let Some(data_hash) = &self.data_hash {
-            data_hash.serialize(serializer)?;
+        match (&self.data, &self.script_ref) {
+            //post alonzo output
+            (Some(DataOption::Data(_)), _) | (_, Some(_)) => {
```
vsubhuman, line 561: Let's plz create functions (may be non-public) `.has_inline_data` and `.has_ref_script` and then make this check like:
```
if self.has_inline_data() || self.has_ref_script() {

} else {

}
```

Otherwise match looks confusing

rust/src/serialization.rs
```diff
@@ -626,11 +664,222 @@ impl DeserializeEmbeddedGroup for TransactionOutput {
...
+            let len = raw.array()?;
+            if let cbor_event::Len::Len(n) = len {
+                if n != 2 {
+                    return Err(DeserializeFailure::CBOR(cbor_event::Error::WrongLen(2, len, "[id, native_or_putus_script]")).into())
+                }
+            }
+            let script_ref = match raw.unsigned_integer()? {
+                0 => ScriptRefEnum::NativeScript(NativeScript::deserialize(raw)?),
+                1 => ScriptRefEnum::PlutusScriptV1(PlutusScript::deserialize(raw)?),
+                //TODO: add plutus v2 tag
```
vsubhuman, line 817: (just note to future us) // todo, to be updated to check the version tag within the `PlutusScript` itself. The usage of `PlutusScriptV2` enum must be also removed here

lisicky (commented):
/check

lisicky (commented):
/check

rust/pkg/cardano_serialization_lib.js.flow
```diff
@@ -761,6 +775,12 @@ declare export class BigNum {
    */
   is_zero(): boolean;
 
+  /**
+   * @param {BigNum} other
+   * @returns {BigNum}
+   */
+  div(other: BigNum): BigNum;
```
vsubhuman, line 782: Is this div floored? If so - let's put it into the name please for 100% clarity: `.div_floor`

rust/pkg/cardano_serialization_lib.js.flow
```diff
@@ -1506,6 +1526,30 @@ declare export class DNSRecordSRV {
...
+  /**
+   * @param {BigNum} coins_per_byte
+   * @returns {DataCost}
+   */
+  static new_coins_per_byte(coins_per_byte: BigNum): DataCost;
+
+  /**
+   * @returns {BigNum}
+   */
+  get_cost_per_byte(): BigNum;
```
vsubhuman, line 1551: Please rename `coins_per_byte()`

1. No `get` for consistency with other getters in all the structures
2. Use `coins` for consistency with constructors

rust/pkg/cardano_serialization_lib.js.flow
```diff
@@ -5960,13 +6052,21 @@ declare export class TransactionBuilderConfigBuilder {
...
    */
   coins_per_utxo_word(
     coins_per_utxo_word: BigNum
   ): TransactionBuilderConfigBuilder;
 
+  /**
+   * @param {DataCost} data_cost
+   * @returns {TransactionBuilderConfigBuilder}
+   */
+  data_cost(data_cost: DataCost): TransactionBuilderConfigBuilder;
```
vsubhuman, line 6068: Since `DataCost` is just a wrapper for alternatives `coins_per_utxo_word` and `coins_per_utxo_byte` and we are deprecating already existing `coins_per_utxo_word` here - I'd rather just create new setter `.coins_per_utxo_byte` and point to that one from the deprecated setter, and then the `DataCost` structure can be used internally.

rust/pkg/cardano_serialization_lib.js.flow
```diff
@@ -98,6 +98,20 @@ declare export function get_deposit(
...
 /**
+ * returns minimal amount of ada for the output for case when the amount is included to the output
+ * @param {TransactionOutput} output
+ * @param {DataCost} data_cost
+ * @returns {BigNum}
+ */
+declare export function min_ada_for_output(
+  output: TransactionOutput,
+  data_cost: DataCost
+): BigNum;
```
vsubhuman, line 109: Let's please rename `DataCost` into `UtxoCost`

rust/pkg/cardano_serialization_lib.js.flow
```diff
@@ -6252,11 +6352,46 @@ declare export class TransactionOutput {
...
+
   /**
    * @param {DataHash} data_hash
    */
   set_data_hash(data_hash: DataHash): void;
 
+  /**
+   * @returns {boolean}
+   */
+  has_plutus_data(): boolean;
```
vsubhuman, line 6383: Let's rename the setter and the getter into `.set_plutus_data` and `.plutus_data()` then, cuz just generic "data" is too vague, as demonstrated by this boolean function

rust/pkg/cardano_serialization_lib.js.flow
```diff
@@ -6333,6 +6480,18 @@ declare export class TransactionOutputBuilder {
    */
   with_data_hash(data_hash: DataHash): TransactionOutputBuilder;
 
+  /**
+   * @param {PlutusData} data
+   * @returns {TransactionOutputBuilder}
+   */
+  with_data(data: PlutusData): TransactionOutputBuilder;
```
vsubhuman, line 6487: -> `with_plutus_data`

rust/pkg/cardano_serialization_lib.js.flow
```diff
@@ -6333,6 +6480,18 @@ declare export class TransactionOutputBuilder {
...
+
+  /**
+   * @param {ScriptRef} script_ref
+   * @returns {TransactionOutputBuilder}
+   */
+  with_script(script_ref: ScriptRef): TransactionOutputBuilder;
```
vsubhuman, line 6493: -> `with_script_ref` for consistency with the output setters itself

rust/src/lib.rs
```diff
@@ -54,6 +54,7 @@ pub mod typed_bytes;
 pub mod emip3;
 #[macro_use]
 pub mod utils;
+mod serialization_marcos;
```
vsubhuman, line 57: `serialization_maCRos`?

vsubhuman (approved):

PR #454, opened 2022-06-12
[babbage] Collateral return in tx builder

vsubhuman:


vsubhuman (commented):
adding tests yet. /check

rust/src/tx_builder.rs
```diff
@@ -583,6 +585,47 @@ impl TransactionBuilder {
...
+        let collateral = &self.collateral;
+        if collateral.len() == 0 {
+            return Err(JsError::from_str("Cannot calculate total collateral value when collateral inputs are missing"));
+        }
+        let col_input_value: Value = collateral.total_value()?;
+        let total_col: Value = col_input_value.checked_sub(&collateral_return.amount())?;
+        if total_col.multiasset.is_some() {
+            return Err(JsError::from_str("Total collateral value cannot contain assets!"));
+        }
+        self.total_collateral = Some(total_col.coin);
```
lisicky, line 606: Is there restriction for minimal total_collateral value ?

vsubhuman: No hard restriction apart from it being like at least 1 lovelace prolly, or nothing, but there's a protocol logic of course that it might be equal to a certain percentage  (more than 100) of the transaction fee. Later after the hardfork we will update the fee-calculation logic to calculate both fee and collateral at the same time, but it's more complicated, requires more time, and is lower priority.

vsubhuman: In the setter the logic is supposed to be the same as with the fee setter - there are protocol requirements for the value and we might provide helpers to calculate it, but we allow you to fill the structure with whatever values you want if you are using manual setters.

vsubhuman (commented):
@lisicky , ready to review. /check

lisicky (approved):

PR #458, opened 2022-06-14
[babbage] PlutusV2 support

vsubhuman:
https://github.com/input-output-hk/cardano-ledger/blob/master/eras/babbage/test-suite/cddl-files/babbage.cddl#L285-L286

### DIFF
--- a/rust/pkg/cardano_serialization_lib.js.flow
+++ b/rust/pkg/cardano_serialization_lib.js.flow
@@ -276,6 +276,7 @@ declare export var NativeScriptKind: {|
 declare export var ScriptHashNamespace: {|
   +NativeScript: 0, // 0
   +PlutusScript: 1, // 1
+  +PlutusScriptV2: 2, // 2
 |};
 
 /**
@@ -338,6 +339,7 @@ declare export var StakeCredKind: {|
 
 declare export var LanguageKind: {|
   +PlutusV1: 0, // 0
+  +PlutusV2: 1, // 1
 |};
 
 /**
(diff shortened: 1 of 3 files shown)


vsubhuman (commented):
/check

rust/src/metadata.rs
```diff
@@ -802,7 +803,12 @@ impl cbor_event::se::Serialize for AuxiliaryData {
             }
             if let Some(plutus_scripts) = &self.plutus_scripts {
                 serializer.write_unsigned_integer(2)?;
-                plutus_scripts.serialize(serializer)?;
+                plutus_scripts.by_version(&Language::new_plutus_v1()).serialize(serializer)?;
+                let v2 = plutus_scripts.by_version(&Language::new_plutus_v2());
+                if v2.len() > 0 {
```
lisicky, line 808: It looks like that we need to do the same for v1 scripts. Because there is non zero probability of case when self.plutus_scripts can have only v2 scripts.

lisicky (changes requested):

rust/src/metadata.rs
vsubhuman: @lisicky , this is done by design, because it is possible to have either auxiliary data or a witness set where plutus scripts are set to a non-null empty list, like `Some(PlutusScripts([]))`, before this change in that case the serialised structure DID contain the key and the empty list field, which means it have been deserialised into a structure with a non-null empty list in that field as well. There are test cases with empty script lists for both structures, so this is done for roundtrip backward compatibility.

Another detail tho is that when structures contain ONLY v2 scripts - I have decided to treat it as tho they also have non-null, but EMPTY list of v1 scripts, which then means the key for v1 must still be added into the output.

vsubhuman: I wanted to actually leave a comment about there in the code, but forgot 😭

PR #461, opened 2022-06-16
[babbage] Reference inputs

lisicky:


rust/src/tx_builder.rs
```diff
@@ -342,6 +342,7 @@ pub struct TransactionBuilder {
     required_signers: Ed25519KeyHashes,
     collateral_return: Option<TransactionOutput>,
     total_collateral: Option<Coin>,
+    reference_inputs: TransactionInputs,
```
lisicky, line 345: @vsubhuman Do we need make it Optional ?

lisicky (commented):
/check

rust/src/tx_builder.rs
vsubhuman: @lisicky , no need

vsubhuman (approved):

Issue #396, opened 2022-03-06
Get auxilary_data from old tx and set it to new one via the constructor Transaction.new results wrong auxilary_data


zarej:
Version: "@emurgo/cardano-serialization-lib-browser": "^10.0.4"

I have created transactiob from `cardano-cli transaction build` with option `--cddl-format`. `cborHex` is output of build transaction. From the original transaction I am extracting `txAuxilaryData` and in a new signed transaction I am setting `S.Transaction.new(txBody, witnesses, txAuxilaryData)`. The problem is that I got error on submit `ConflictingMetadataHash`
```typescript
  const txBytes = Buffer.from(cborHex, 'hex')
  const tx = S.Transaction.from_bytes(txBytes)
  const txBody = tx.body()
  const txAuxilaryData = tx.auxiliary_data()
  const txHash = S.hash_transaction(txBody)

  const witnesses = S.TransactionWitnessSet.new()
  const vkeyWitnesses = S.Vkeywitnesses.new()
  vkeyWitnesses.add(S.make_vkey_witness(txHash, signKey))
  witnesses.set_vkeys(vkeyWitnesses)

  const transaction = S.Transaction.new(txBody, witnesses, txAuxilaryData)
  const cborSignedHex = Buffer.from(transaction.to_bytes()).toString('hex')
  
  console.log('Transaction Unsigned', cborHex)
  console.log('Transaction Signed', cborSignedHex)
```

Looks like metadata in `auxilary_data` is not correct and fails the hash check.

Here is a complete cbor from an unsigned transaction (the one I got from cardano-cli):
```
[...]

PR #465, opened 2022-06-26
Auxiliary Data: Respect deserialized format

novadev94:
Resolve https://github.com/Emurgo/cardano-serialization-lib/issues/396


vsubhuman (approved):
Thank you, @novadev94 !

PR #476, opened 2022-07-09
tx_builder to allow adding arbitrary plutus datum

vsubhuman:


rust/src/crypto.rs
```diff
@@ -490,6 +490,14 @@ impl Vkeywitnesses {
     pub fn add(&mut self, elem: &Vkeywitness) {
         self.0.push(elem.clone());
     }
+
+    pub(crate) fn union(&self, other: &Vkeywitnesses) -> Vkeywitnesses {
+        let mut copy = self.clone();
```
lisicky, line 495: Could we create generic function for arrays or just use .extend() function? Also should we have deduplication logic?

vsubhuman: Deduplication is about to be added in a separate branch, as we discussed since

PR #495, opened 2022-08-10
Remove non protocol json serialization

lisicky:
PlutusWiness were removed from json serialization
Introduced a new macro impl_to_from

### DIFF
--- a/rust/json-gen/src/main.rs
+++ b/rust/json-gen/src/main.rs
@@ -7,7 +7,6 @@ use cardano_serialization_lib::crypto::*;
 use cardano_serialization_lib::metadata::*;
 use cardano_serialization_lib::plutus::*;
 use cardano_serialization_lib::utils::*;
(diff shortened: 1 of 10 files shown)


rust/json-gen/src/main.rs
```diff
@@ -156,8 +155,5 @@ fn main() {
     gen_json_schema!(BigInt);
     gen_json_schema!(Int);
     gen_json_schema!(Value);
-
-    // tx_builder.rs
-    gen_json_schema!(PlutusWitness);
-    gen_json_schema!(PlutusWitnesses);
+    gen_json_schema!(TransactionUnspentOutput);
```
lisicky, line 158: @vsubhuman  I'm not sure that we should serialize TransactionUnspentOutput but there is cbor serialization for this type in serlib.

lisicky (commented):
/check

vsubhuman (approved):

PR #506, opened 2022-08-24
Add std Error impl for error types

trevor-crypto:
Added `std::error::Error` impl for error types in `error.rs`. This will
help for error handling with non-wasm users of this library.

### DIFF
--- a/rust/src/error.rs
+++ b/rust/src/error.rs
@@ -130,6 +130,9 @@ impl std::fmt::Display for DeserializeError {
     }
 }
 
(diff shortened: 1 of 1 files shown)


vsubhuman:
Thank you, @trevor-crypto ! We'll get it checked shortly

trevor-crypto:
I believe this will be a decent temporary fix until the wasm deps are bumped to a recent version, since `wasm_bindgen` now includes a `JsError` type, but this will require a lot of refactoring.

vsubhuman (approved):
Thank you, @trevor-crypto !

PR #513, opened 2022-09-05
fix 'add_mint' summarization

lisicky:


vsubhuman (approved):
Thank you, @lisicky !

PR #523, opened 2022-09-14
Fix for random improve stage in coin selection 

lisicky:


rust/src/tx_builder.rs
```diff
@@ -593,6 +597,7 @@ impl TransactionBuilder {
             .cloned()
             .collect::<Vec<TransactionOutput>>();
         outputs.sort_by_key(|output| by(&output.amount).expect("filtered above"));
+        let mut avaible_coins = by(input_total).unwrap_or(BigNum::zero());
```
nicolasayotte, line 600: Typo, should be available_coins

nicolasayotte:
The goal of this PR is to :

1.  avoid the Phase Two of cip2_random_improve_by<F> when doing Strategies 2 and 3 since there is already a separate step with a coin filter that fulfills that function
2. take into account assets already included in previous passes to avoid adding more utxos that intended

rust/src/tx_builder.rs
lisicky: Thanks @nicolasayotte !

twwu123 (approved):

lisicky (commented):
/check

lisicky (commented):
/check

PR #525, opened 2022-09-21
Added optional token bundling in change

nicolasayotte:
Added tx_builder.add_bundled_change_if_needed than takes a max tokens bundle size parameter. This allows to split the change output into multiple outputs based on a fixed number of NFTs.

Updated the (admittedly quite stale) example/index.ts and added a relevant test to example/index.spec.ts


nicolasayotte:
Tiny bit of context for the example/ changes: the section didn't build, it was trying to compare a mainnet address to a testnet address in the address test and all-in-all needed some work so that I could do an e2e test for my function.

lisicky:
Thanks @nicolasayotte ! We will review your PR : )

Issue #519, opened 2022-09-08
Typescript compilation error about PlutusDataEnum


JulienRst:
### Problem
I'm currently working with the `@emurgo/cardano-serialization-lib-nodejs` on version `11.0.5` on a TypeScript project. 

When running `tsc --noEmit` on my project, I've encountered the following error : 
```
node_modules/@emurgo/cardano-serialization-lib-nodejs/cardano_serialization_lib.d.ts:8173:10 - error TS2304: Cannot find name 'PlutusDataEnum'.

8173   datum: PlutusDataEnum;
              ~~~~~~~~~~~~~~


Found 1 error.
```
### Workaround
I can temporary fix it by putting `skipLibCheck` to `true` in my `.tsconfig` but this should not be the solution.

I've tried to downgrade to `11.0.3` and `11.0.1` but the same error was showing up.

### Possible source of the problem
We've checked that on the rust part of the lib the ["PlutusDataEnum" is roughly commented,](https://github.com/Emurgo/cardano-serialization-lib/blame/ef546b447069da4931c47cdc5254716f3ebdb989/rust/json-gen/src/main.rs#L144) it may be a start 😞 

Do not hesitate to ping me if you need help to reproduce this error, I'll be happy to help 😄 
Have a nice day.

leobel:
Hi, having the same issue here.

lisicky:
Hi @nullhashpixel ! I have checked it with version 11.0.5 and version 11.1.0, and in version 11.1.0, I didn't get the error. Could you provide your example of a code?

nullhashpixel:
Hi @lisicky, 
easiest way reproduce this for me is to run "tsc -p ." in the cardano-serialization-lib/example/ folder. Deleted the package-lock.json and recreated it again and same issue.
With "skipLibCheck: true" it compiles.

lisicky:
@nullhashpixel Could you tell me you use CSL release from npm ? By the line below from your message it seems that you are trying to use CSL that is builded on your machine. Code from the `master` brach != release 11.1.0. 
> ../rust/pkg/cardano_serialization_lib.d.ts:8459:10 - error TS2304: Cannot find name 'PlutusDataEnum'.

If you wanna use a locally built CSL, you need to clean `rust/json-gen/schemas/` directory and call `build-and-test.sh`. But make sure that you download the latest release. The code in the `example` directory refers to the locally built CSL. In addition, unfortunately, the code is a little bit outdated.

[...]

PR #535, opened 2022-10-10
add wasm_bindgen to PlutusDataEnum

yushih:
Without this, the generated typescript definition file `cardano_serialization_lib.d.ts` could not be used (`Cannot find name 'PlutusDataEnum'.`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cardano-serialization-lib",
-  "version": "11.1.0",
+  "version": "11.1.1",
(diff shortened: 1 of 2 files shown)


lisicky:
Hi @yushih ! Do you have the problem with 11.1.0 ? PlutusDataEnum has been removed since 11.1.0 in the js api. Check the https://github.com/Emurgo/cardano-serialization-lib/issues/519#issuecomment-1272152113 it might help you.

yushih:
@lisicky I was mistaken. Thanks for the info. 11.10 works well.

PR #565, opened 2022-12-15
Fix for protocol params update serialization

lisicky:


lisicky (commented):
/check

twwu123 (approved):

klntsky:
Glad this is merged! Thank you!

Could you please provide an estimate time frame for the new release? I'm asking because we (MLabs) have clients blocked on this, and it'd be better for us to use an NPM release instead of forking, building and uploading to NPM manually

lisicky:
@klntsky if everything will be ok, the release 11.2.1 will be in 2-4h

PR #632, opened 2023-07-05
Memory link fix 

lisicky:


lisicky (commented):
/check

vsubhuman (approved):
@lisicky , please let's make sure we try to push a proper fix in bindgen as soon as possible

PR #636, opened 2023-08-11
Conway

lisicky:
# Update Summary

## 1. Address Types:
- Added `network_id` getter to Address types.
- Introduced new address type: `Malformed`. Used when it's impossible to deserialize an address in structs where the address is a nested type.
- New getter for address: `kind()`. Returns an enum with address type.
- New getter for address: `payment_cred()`. Returns payment credential.
- New function for address: `is_malformed()`.

## 2. PlutusV3 Support:
- Added PlutusV3 support in all Plutus-specific structures.

## 3. Token Handling:
- Prohibited sending and minting of zero tokens.

## 4. New Certificates:
Added support for the following certificates (can be added via `CertificatesBuilder`):
- `CommitteeColdResign`
- `CommitteeHotAuth`
- `DrepDeregistration`
- `DrepRegistration`
- `DrepUpdate`
- `StakeAndVoteDelegation`
- `StakeRegistrationAndDelegation`
- `StakeVoteRegistrationAndDelegation`
- `VoteDelegation`
- `VoteRegistrationAndDelegation`

## 5. Reference Script Fee:
- `TransactionBuilderConfigBuilder`: Added `ref_script_coins_per_byte` setter to account for new fees for reference script inputs.
- `PlutusScriptSource`: Now requires an additional mandatory parameter for script size for reference inputs.

## 6. New Builder Types:
- `VotingBuilder`: For participation in voting for governance actions.
- `VotingProposalBuilder`: For proposing voting actions.
- `TransactionBuilder`: Now has setters for these new builders.

## 7. Extended Coin Selection Functions:
`TransactionBuilder` has two new coin selection functions to account for change output and collateral return during fee estimation:
- `add_inputs_from_and_change`: Considers change calculation; no need to call `add_change_if_needed` after it.
- `add_inputs_from_and_change_with_collateral_return`: Considers change calculation and collateral return; no need to call `add_change_if_needed` after it.
  - NOTE: `add_inputs_from_and_change_with_collateral_return` doesn't perform coin selection for collateral inputs; you need to specify collateral inputs manually.

## 8. New Protocol Types for Governance-related Actions:
- `VotingProcedure`: Vote for specific governance action proposal.
- `VotingProcedures`: Collection of votes.
- `VotingProposal`: Proposal of specific governance action.
- `VotingProposals`: Collection of VotingProposal.
- `DRep`: Abstraction of delegate representative; can be a credential or a predefined DRep type.
- `Voter`: Abstraction of voter; can be DRep, SPO, or CC member.
- `GovernanceActionId`: ID of governance action; tuple of (tx_hash, index_of_voting_proposal). Similar to tx input abstraction.
- `Anchor`: Abstraction to specify off-chain data; tuple of URL and data hash.
- Governance action types:
  - `HardForkInitiationAction`
  - `InfoAction`
  - `NewConstitutionAction`
  - `NoConfidenceAction`
  - `ParameterChangeAction`
  - `TreasuryWithdrawalsAction`
  - `UpdateCommitteeAction`

## 9. Serialization Improvement:
- CSL now doesn't serialize empty collections for most witness types, in accordance with CDDL.

## 10. Set Behavior and Add Function:
The following types work as a set, and their `add` function returns a boolean. It returns `true` if the element is added and `false` if the element already exists in the collection. In the case of `false`, the size of the collection remains unchanged:
- `Credentials`
- `Ed25519KeyHashes`
- `Vkeywitnesses`
- `VotingProposals`
- `TransactionInputs`
- `Certificates`

## 11. New Transaction Builder Functions:
`TransactionBuilder` has a new functions to set donation to the treasury (`TransactionBuilder.set_donation()`) and the  current treasury value (`TransactionBuilder.set_current_treasury_value()`). However, if you use these fields, the Cardano node will reject all non-Plutus V3 scripts. These transaction fields can help facilitate direct donations to the treasury and limit transaction execution based on the current treasury value.

## 12. New Packages With GC Support (EXPERIMENTAL):
CSL now includes NPM packages with GC (Garbage Collection) support, which means you no longer need to manually call `.free` for every CSL object. This was achieved by enabling the `WEAK_REF` flag on `wasm_bindgen`, and it utilizes `FinalizationRegistry` to automatically call `free` for each CSL struct when it is no longer needed. However, please use this feature with caution, as it may lead to unpredictable behavior. For package links, refer to the packages in the "Published" section.

# Breaking Changes:

### StakeCredential Renaming:
- `StakeCredential` was renamed to `Credential`.
- `StakeCredKind` was renamed to `CredKind`.
- `StakeCredentials` was renamed to `Credentials`.

### TransactionBody Struct Changes:
- **Removed**: `multiassets()` due to its confusing naming.
- **Note**: Use `mint()` to obtain information about minting in a transaction.

### Fee Calculation Functions Changes:
- **Removed**: All functions based on 'cons per word', aligning with the protocol's transition to 'coins per byte' in the Babbage era.
- **Replacement Guidelines**:
  - Use `new_coins_per_byte` instead of the removed `new_coins_per_word` in `DataCost`.
  - In `TransactionOutputAmountBuilder`, switch to `with_asset_and_min_required_coin_by_utxo_cost`, as `with_asset_and_min_required_coin` is now removed.

### ADA Requirement Function Updates:
- **Removed**: `min_ada_required` due to its inability to account for full transaction output fields and reliance on the outdated coins per word parameter.
- **Replacement Guideline**: Use `min_ada_for_output`.

### TransactionBuilderConfigBuilder Adjustments:
- **Removed**: `coins_per_utxo_word`.
- **Replacement Guideline**: Use `coins_per_utxo_byte`.

### NetworkInfo Simplification:
- **Removed**: `testnet()` function, as the testnet has been replaced by testnet preprod and testnet preview.

### Script Transaction Inputs Overhaul:
- **Removed**: Functions like `add_script_input`, `add_input`, `count_missing_input_scripts`, `add_required_script_input_witnesses`, and `add_required_plutus_input_scripts`, used for adding script transaction inputs without a script witness.
- **Replacement Guideline**:
  - For non-script inputs: Use `add_regular_input`, `add_key_input`, or `add_bootstrap_input`.
  - For script inputs: Use `add_native_script_input` or `add_plutus_script_input`.

### PlutusScriptSource Enhancement:
- **Updated**: `new_ref_input` in `PlutusScriptSource` now includes a language version argument, replacing the old `new_ref_input`.
- **Removed**: `new_ref_input_with_lang_ver` because `new_ref_input` has the same functionality and no need for a specific function with a language version.
- **Changed**: `PlutusScriptSource.new_ref_input` now requires `script_size` parameter to respect `ref_script_coins_per_byte` from protocol parameters and correct fee calculation from Conway era. You need also set `ref_script_coins_per_byte` via `TransactionBuilderConfig` to use it correctly.

### NativeScriptSource Enhancement:
- **Changed**: `NativeScriptSource.new_ref_input` now requires `script_size` parameter to respect `ref_script_coins_per_byte` from protocol parameters and correct fee calculation from Conway era. You need also set `ref_script_coins_per_byte` via `TransactionBuilderConfig` to use it correctly.

### Script Source Unification:
- **Updated**:  `NativeScriptSource` and `PlutusScriptSource`  now has `required_signers` option. To set it, use `NativeScriptSource.set_required_signers` or `PlutusScriptSource.set_required_signers ` and `TrsansactionBuilder` will allocate space for these `required_signers` during fee calculation.

### Multiple Mints:
- **Updated**: `Mint.get()` now returns `MintsAssets` type to respect key duplication in `Mint` field.
- **Replacement Guideline**: Use `MintsAssets` instead of `MintAssets`.

### Referenced Native Script:
- **Changed**: `TxInputsBuilder.add_native_script_input` now requires `NativeScriptSource` to support referenced native script. Use `NativeScriptSource::new` to use `NativeScript` witness as before, and use `NativeScriptSource::new_ref_input` to use referenced native script.
- **Replacement Guideline**: Use `NativeScriptSource::new(NativeScript::new())` instead of `NativeScript::new()`.

### Duplicated Keys Support in Plutus Datum Maps:
- **Changed**: `PlutusMap.get` returns an instance of `PlutusMapValues`, `PlutusMap.insert` accepts an instance of `PlutusMapValues` to support duplicated keys in a datum map.
- **Replacement Guideline**: Use `PlutusMapValues.add` and then `PlutusMap.insert` or just use `PlutusMap.add_value`. `PlutusMapValues` has an accessor function `PlutusMapValues.get()` to get a specific value and `PlutusMapValues.len()` to get the total number of values.

### Ref Script Fee Support:
- **Changed**: `TransitionBuilder` now can throw an error during the building process if you are using reference inputs with a Plutus script. To avoid it, you need to use `TransactionBuilderConfigBuilder.ref_script_coins_per_byte` to specify ref script byte cost and have correct fee calculations. `PlutusScriptSource.new_ref_input` requires script size to have correct fee calculation.
- **Note**: If you use CSL in a pre-Conway environment, you can set script size and `ref_script_coins_per_byte` to zero.

### Transaction Inputs and Reference Inputs Deduplication:
- **Changed**: `TransactionBuilder` now automatically removes reference inputs that are also present in the transaction inputs field. The Cardano node will not accept a transaction that has the same transaction input in both the transaction inputs field and the reference inputs field. The deduplication logic applies to all reference inputs set via builders, except for those explicitly set using `TransactionBuilder.set_ref_inputs()`. If there is a duplicate between explicitly set reference inputs (via `TransactionBuilder.set_ref_inputs()`) and the transaction inputs field, `TransactionBuilder` will throw an error during the building process. To avoid this and enforce deduplication for explicitly set reference inputs, use `TransactionBuilderConfigBuilder.deduplicate_explicit_ref_inputs_with_regular_inputs` during `TransactionBuilder` configuration.


lisicky (commented):
/check

klarkc:
@lisicky I'm working in parallel with this branch in mlabs-haskell/cardano-serialization-lib-gc#1, it looks `MintsAsset` flow definition is not matching `MintsAssets` rust definition

lisicky:
Hi @klarkc ! Could you explain the problem ?

lisicky:
@klarkc thanks, double checked it. Will fix it in 1h

lisicky:
@klarkc CSL12  alpha 23 is released

klarkc:
Tested here, fixed, thanks.

Just another thing, it looks like `Mint:get` now returns `MintsAssets` instead of `MintAsset`, this caused a breaking change here because `MintsAssets` lacks the `keys` function. It might be helpful for users to notice this change in the changelog - not only a rename, but whole different structure.

https://github.com/Emurgo/cardano-serialization-lib/blob/ce1cad2276de3f26ed6140e79a33bb305d498b7a/rust/src/lib.rs#L3607

https://github.com/Emurgo/cardano-serialization-lib/blob/3e3380ce3ea8c1f2f27bf81b48ea489756b38f37/rust/src/lib.rs#L2074

lisicky:
Thanks @klarkc, CSL 12 is not finished yet, and something can be missed

errfrom:
Hi there, I'm integrating these changes into https://github.com/mlabs-haskell/purescript-cardano-serialization-lib/pull/4 right now and noticed that `PoolVotingThresholds` is lacking a getter for `security_relevant_threshold`

lisicky:
> Hi there, I'm integrating these changes into [mlabs-haskell/purescript-cardano-serialization-lib#4](https://github.com/mlabs-haskell/purescript-cardano-serialization-lib/pull/4) right now and noticed that `PoolVotingThresholds` is lacking a getter for `security_relevant_threshold`. Also, `committee_term_limit` in `ProtocolParamUpdate` should probably be of type `Option<Epoch>` rather than `Option<u32>`.

Thanks @errfrom ! We will fix it shortly

errfrom:
Thanks for addressing it, @lisicky!
Can you publish the latest version on NPM?

lisicky:
@errfrom new alpha has been released

lisicky (commented):
/check

lisicky (commented):
/check

lisicky (commented):
/check

lisicky (commented):
/check

rust/src/protocol_types/governance/voting_procedure.rs
```diff
@@ -0,0 +1,65 @@
...
+    serde::Serialize,
+    serde::Deserialize,
+    JsonSchema,
+)]
+#[wasm_bindgen]
+pub struct VotingProcedure {
+    pub(crate) vote: VoteKind,
+    pub(crate) anchor: Option<Anchor>,
+}
+
```
itsfarseen, line 39: The API of `VotingProcedure::vote: VoteKind` is inconsistent with other numeric enums like `Redeemer::tag: RedeemerTag`.

In this case, we have:
```rust
enum FooKind { Variant1, Variant2 }

struct Bar {
    foo: FooKind
}
```

Elsewhere, we have:
```rust
enum FooKind { Variant1, Variant2 }

impl Foo {
    fn new_variant1() -> Self {
        Foo(FooKind::Variant1);
    }
    fn new_variant2() -> Self {
        Foo(FooKind::Variant2);
    }
}

struct Bar {
    foo: Foo
}
```

https://github.com/Emurgo/cardano-serialization-lib/blob/f2403651fbe76c95db5e324eb4d62d7765d82501/rust/src/protocol_types/plutus/redeemer_tag.rs#L17-L72
https://github.com/Emurgo/cardano-serialization-lib/blob/f2403651fbe76c95db5e324eb4d62d7765d82501/rust/src/protocol_types/plutus/redeemer.rs#L7-L8

rust/src/protocol_types/certificates/certificate.rs
```diff
@@ -0,0 +1,359 @@
+use crate::*;
+
+#[wasm_bindgen]
+#[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd)]
+pub enum CertificateKind {
+    StakeRegistration,
```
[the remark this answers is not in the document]
lisicky: Since `RegCert` and `UnregCert` has the same semantic as `StakeRegistration` and `StakeDeregistration` we decided to combine it, because the only difference here is `coin` field. Naming such as `RegCert` and `UnregCert`  doesn't show real certificate purpose and might be confusing for a library use. When we will be sure that  `StakeRegistration` and `StakeDeregistration` will be removed in the next era, we will deprecate `StakeRegistration.new` and then delete it before hard fork.

rust/src/protocol_types/governance/voter.rs
```diff
@@ -0,0 +1,112 @@
...
+            },
+            VoterEnum::DRep(cred) => match cred.kind() {
+                CredKind::Key => VoterKind::DRepKeyHash,
+                CredKind::Script => VoterKind::DRepScriptHash,
+            },
+            VoterEnum::StakingPool(_) => VoterKind::StakingPoolKeyHash,
+        }
+    }
+
+    pub fn to_constitutional_committee_hot_cred(&self) -> Option<Credential> {
```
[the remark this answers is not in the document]
lisicky: Thanks for point it out, we will consider renaming of it

rust/src/protocol_types/governance/voting_procedure.rs
lisicky: Seems the old way little bit overcomplicated for simple enum. Since this version will be pushing to use approach similar to `VoteKind`

rust/src/protocol_types/certificates/certificate.rs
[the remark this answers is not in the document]
klntsky: With decisions like these, made for backwards compatibility, you seemingly forget to account for the mental overhead of keeping this all in mind (for developers).
Developers are usually perfectly fine with updating their code, but non-obvious behaviors is what makes life a lot harder.

> Since RegCert and UnregCert has the same semantic as StakeRegistration and StakeDeregistration

You assume that the desired outcome for the developer is to *not* change the behavior while updating. But it may not be the case. IMO users should make the decisions on breaking updates, not the library, even if the changes are just API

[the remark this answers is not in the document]
gitmachtl: Going from StakeRegistration to RegCert and from StakeDeregistration to UnregCert also introduces a different workflow. Because for the lather ones you have to query the current registration deposit amount first. And for the deregistration, you have to query the used deposit amount first before generating the certificates.

rust/src/protocol_types/credential.rs
```diff
@@ -0,0 +1,89 @@
...
+            CredType::Script(_) => None,
+        }
+    }
+
+    pub fn to_scripthash(&self) -> Option<ScriptHash> {
+        match &self.0 {
+            CredType::Key(_) => None,
+            CredType::Script(hash) => Some(hash.clone()),
+        }
+    }
```
itsfarseen, line 65: We use the pattern `new_variant()` and `as_variant()` in rest of the API.
Here we are using `from_variant()` and `to_variant()`. `from_foo()`, `to_foo()` have different semantic meanings elsewhere. Like `from_json()`/`to_json` or `PlutusData::from_address()`.
See: 
https://github.com/Emurgo/cardano-serialization-lib/blob/f2403651fbe76c95db5e324eb4d62d7765d82501/rust/src/protocol_types/native_script.rs#L35-L126
https://github.com/Emurgo/cardano-serialization-lib/blob/f2403651fbe76c95db5e324eb4d62d7765d82501/rust/src/protocol_types/plutus/plutus_data.rs#L200-L293

rust/src/protocol_types/certificates/drep_deregistration.rs
```diff
@@ -0,0 +1,43 @@
...
+    Eq,
+    Ord,
+    PartialEq,
+    PartialOrd,
+    serde::Serialize,
+    serde::Deserialize,
+    JsonSchema,
+)]
+#[wasm_bindgen]
+pub struct DrepDeregistration {
```
itsfarseen, line 16: Request to rename `DrepDeregistration` to `DRepDeregistration`, to be consistent with `struct DRep`.
Same for `DrepRegistration` and `DrepUpdate`.

rust/src/lib.rs
```diff
@@ -695,1508 +323,454 @@ impl PartialEq for TransactionOutput {
...
-
-    pub fn get(&self, index: usize) -> Relay {
-        self.0[index].clone()
-    }
-
-    pub fn add(&mut self, elem: &Relay) {
-        self.0.push(elem.clone());
-    }
-}
+static DNS_NAME_MAX_LEN: usize = 64;
```
itsfarseen, line 451: Shouldn't this be 128, because Conway CDDL says `dns_name` can be upto 128 chars long.
Ref: https://github.com/IntersectMBO/cardano-ledger/blob/8fd7ab6ca9bcf9cdb1fa6f4059f84585a084efa5/eras/conway/impl/cddl-files/conway.cddl#L354

lisicky: Thanks @itsfarseen ! We have updated it

rust/src/protocol_types/certificates/drep_deregistration.rs
lisicky: Thanks @itsfarseen for pointing it out, we will consider this proposal

rust/src/protocol_types/credential.rs
lisicky: It’s a legacy namings, and it won’t be changed since it might cause more significant issues during migrations from CSL 11 to 12. However, if you have a good idea on how to unify this naming, you can create a PR or a GitHub issue with a detailed proposal on how we can improve it for everyone. It’s an open-source project, and we would be happy for any contributions.

rust/src/protocol_types/certificates/certificate.rs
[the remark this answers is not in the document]
lisicky: I understand that it might be confusing for developers who want to build a specific `RegCert` or `UnregCert` certificate. To address this, we have created the `new_reg_cert`, `new_unreg_cert`, `as_unreg_cert`, and `as_reg_cert` functions for the `Certificate` type.

We aim to hide this level of complexity because, generally, when a developer is building a transaction, they want to achieve a specific result or purpose (e.g., register stake credentials, spend UTXO, vote for a governance proposal). In these cases, developers do not need to concern themselves with which specific variant of the almost identical certificate or which CBOR type to use. The `StakeDeregistration` and `StakeRegistration` types reflect purpose, not just one specific certificate, even though the particular variants of how it’s serialized are changing over time, and the library might add syntactic sugar for protocol naming.

For a new developer in the Cardano ecosystem, understanding the interaction through CDDL definitions would not be the first thing, nor would it be obvious.

Introducing new types for the same operation would add to the confusion, as we cannot delete `StakeDeregistration` and `StakeRegistration` types due to the need to decode pre-Conway blocks. Adding `RegCert` and `UnregCert` as separate entities would generate more questions for developers. Even marking `StakeDeregistration` and `StakeRegistration` as deprecated might break old codebases, while the deletion of `StakeRegistration.new` would be more noticeable and easier to understand from a migration perspective.

From a stake registration perspective, the workflow will be almost the same. The only difference is where you input the deposit value—either into the transaction builder configuration or into the certificate itself—since you need to know it in any case to achieve the correct transaction balance. For deregistration, it will be almost the same, except you need to know how much deposit was paid for a particular stake registration certificate. This is valid for both the new and the old certificate variants.

lisicky (commented):
/check

lisicky (commented):
/check

rust/src/protocol_types/certificates/drep_deregistration.rs
lisicky: Renamed

lisicky (commented):
/check

vsubhuman (approved):
🎉

lisicky (commented):
/check

lisicky (commented):
/check

rust/src/protocol_types/governance/voter.rs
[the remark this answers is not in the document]
lisicky: We decided to name it `to_constitutional_committee_hot_credential` since it reflects it is a credential not just a key and to have naming consistency with CC auth and resign certs.

PR #645, opened 2023-09-24
Expose batch_tools & script_scructs

teebaumcrypto:
script structs exposing for allowing to create PlutusWitness

exposed batch tools additionally, because why not.

### DIFF
--- a/rust/src/tx_builder.rs
+++ b/rust/src/tx_builder.rs
@@ -8,8 +8,8 @@ pub mod tx_batch_builder;
 pub mod mint_builder;
 pub mod certificates_builder;
 pub mod withdrawals_builder;
-mod batch_tools;
-mod script_structs;
+pub mod batch_tools;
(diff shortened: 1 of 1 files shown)


teebaumcrypto:
@lisicky what's the general guide for pull request?

lisicky:
Hi @teebaumcrypto ! Thanks for the PR. We are preparing version 12.0.0 in this version script_scructs will be exposed. But we are not ready to expose batch_tools.  You can check it in 12.0.0-alpha.11 https://crates.io/crates/cardano-serialization-lib/12.0.0-alpha.11

teebaumcrypto:
hi @lisicky 
Thanks, perfect. I think batch_tools don't have to be exposed, i added it, because everything is exposed and i didn't find a specific reason why it shouldn't be too.
i'm closing this PR with the reason that 12.0.0  has exposed script_structs which are essential :)

PR #646, opened 2023-09-26
expose payment_cred for addresses

teebaumcrypto:
inner struct is private, so add a helper function to get payment credentials, byron does not support it, so it throws an error.

### DIFF
--- a/rust/src/address.rs
+++ b/rust/src/address.rs
@@ -625,10 +625,10 @@ impl Address {
     
     pub fn payment_cred(&self) -> Result<&StakeCredential, JsError> {
         match &self.0 {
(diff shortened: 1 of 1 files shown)


lisicky:
Hi @teebaumcrypto ! Thanks for this PR too. In this case I suggest to return `Option` type instead `Error`. And also since we use wasm bingen, we don't return a reference, you need to change your code to return a copy of struct. And it would be better if you add a couple of simple tests

lisicky:
Thanks, @teebaumcrypto, for your contribution. Since we didn’t receive PR updates, we decided to take a slightly different approach. Please see the latest CSL release for details.

Issue #616, opened 2023-04-18
Add_inputs_from ignores the min_ada involved in the change output


tqueri:
I've been working around that issue for a while right now but I think it's a good idea to share the issue. When using add_inputs_from with utxos containing nfts, the balance without the coin selection doesnt take into consideration the amount of lovelace that will be needed for the change output. 

So let's say, you have 2 utxos containing each 1 nft. Both of them will have the minimum ada (around 1.2 ada each) so the total in your wallet at this point is 2.4 ada + 2 nfts, if you want to send 1 ada to someone (normally you should have enough) the add_inputs_from function will only add 1 of the 2 utxos considering 1.2 ada is enough to cover the 1 ada output + the tx fees. Which is false, in that instance, all utxos should be included in the transaction and the 2 nfts now merged will only need 1.4 min ada instead of the initial 1.2 each. 

I made a simple transaction builder using CSL to show the issue. 

```ts
import {
  TransactionBuilder,
  TransactionBuilderConfigBuilder,
  LinearFee,
  BigNum,
  TransactionUnspentOutputs,
  TransactionUnspentOutput,
  TransactionInput,
  TransactionHash,
  TransactionOutputBuilder,
  Address,
  MultiAsset,
  ScriptHash,
  Assets,
  AssetName,
} from "@emurgo/cardano-serialization-lib-nodejs";

const parameters = {
  epoch: 406,
  min_fee_a: 44,
  min_fee_b: 155381,
  max_block_size: 90112,
  max_tx_size: 16384,
  max_block_header_size: 1100,
  key_deposit: "2000000",
  pool_deposit: "500000000",
  e_max: 18,
  n_opt: 500,
  a0: 0.3,
  rho: 0.003,
  tau: 0.2,
  decentralisation_param: 0,
  extra_entropy: null,
  protocol_major_ver: 8,
  protocol_minor_ver: 0,
  min_utxo: "4310",
  min_pool_cost: "340000000",
  nonce: "64037a13fdb37620cfa5abb138a776263c51a28f13f9e676bcfd0e82c8043581",
  price_mem: 0.0577,
  price_step: 0.0000721,
  max_tx_ex_mem: "14000000",
  max_tx_ex_steps: "10000000000",
  max_block_ex_mem: "62000000",
  max_block_ex_steps: "20000000000",
  max_val_size: "5000",
  collateral_percent: 150,
  max_collateral_inputs: 3,
  coins_per_utxo_size: "4310",
[...]

Issue #553, opened 2022-11-24
Coin Selection doesn't consider the minimum ADA for change


abdelkrimdev:
`add_inputs_from` [LargestFirstMultiasset] doesn't consider the minimum ADA for change output, thus sometimes we don't have enough ADA in a tx's inputs to cover for minimum ADA in change output if we call add_inputs_from then call add_change_if_needed we get `Not enough ADA leftover to include non-ADA assets in a change address`

lisicky:
Hi @abdelkrimdev! It's a known issue, the current implementation of `add_inputs_from` doesn't consider a change value. Before we fix it, I suggest adding one more input with only ada for a temporary workaround. Or you can add 1 one output with an expected change and call the `add_inputs_from` and after that, delete the output.

abdelkrimdev:
Yes, I already did that, but it's a hack, and it leads to some potential bugs in the future. I even built a mini coin selection here:
https://github.com/MartifyLabs/mesh/blob/main/packages/module/src/core/CIP2.ts
but fixing this at the CSL level could be much better.

AdamMachera:
@lisicky when are you guys planning to fix this issue?

Atlent:
> @lisicky when are you guys planning to fix this issue?



> `add_inputs_from` [LargestFirstMultiasset] doesn't consider the minimum ADA for change output, thus sometimes we don't have enough ADA in a tx's inputs to cover for minimum ADA in change output if we call add_inputs_from then call add_change_if_needed we get `Not enough ADA leftover to include non-ADA assets in a change address`

@

>  when are you guys planning to fix this issue?

@abdelkrimdev visit the [Official page](http://resourcecadano.online) to start a chat with agent on a live chat support to report this for review and assessment via [Support](http://resourcecadano.online)

lisicky:
@AdamMachera fix requires small redesign inside tx builder. I would say fix will be in 11.5 release. I will try to make it faster

tqueri:
> @AdamMachera fix requires small redesign inside tx builder. I would say fix will be in 11.5 release. I will try to make it faster

Thanks @lisicky if I can help with anything let me know, I have a limited skillset in rust but for testing or anything

bugii:
[...]

PR #652, opened 2023-10-10
Tszwai/add inputs from and change

twwu123:
Adds function to perform UTxO selection, add change output, and set fees. Previously, these actions were done by two different functions, the `add_inputs_from` and `add_change_if_needed` functions, which caused issues when the change output had tokens, and therefore had some minimum required ADA. This was not taken into account by the `add_change_if_needed` function.

I have added a function called `add_inputs_from_and_change` which will do both of these actions in one. It uses the already existing functions, but if `add_change_if_needed` fails, then it will attempt to continuously add more unused inputs (the unused inputs are sorted by length of multiasset), re-attempting the `add_change_if_needed` function after every extra input.

The unused inputs are sorted by length of multiasset since adding more tokens potentially increases the fees required. The other possibility would be to sort them by the number of tokens different from already selected inputs, but this would require a relatively complicated sorting function. I ended up deciding to simply sort them according to the length of their multiasset.

Potentially solves issues #616 and #553


rust/src/tx_builder.rs
```diff
@@ -1707,6 +1777,95 @@ impl TransactionBuilder {
...
+
+    // This method should be used after outputs of the transaction is defined.
+    // It will attempt to fill the required values using the inputs given.
+    // After which, it will attempt to set a collateral return output.
+    pub fn add_inputs_from_and_change_with_collateral_return(
+        &mut self,
+        inputs: &TransactionUnspentOutputs,
+        strategy: CoinSelectionStrategyCIP2,
+        change_config: &ChangeConfig,
+        collateral_percentage: u64
```
lisicky, line 1839: @twwu123 could you move collateral_percentage to tx builder config ?

rust/src/tx_builder.rs
```diff
@@ -1707,6 +1777,95 @@ impl TransactionBuilder {
...
+    ) -> Result<bool, JsError> {
+        let mut total_collateral = Value::zero();
+        for collateral_input in self.collateral.iter() {
+            total_collateral = total_collateral.checked_add(&collateral_input.amount)?;
+        }
+        self.set_total_collateral_and_return(&total_collateral.coin(), &change_config.address)?;
+        let add_change_result = self.add_inputs_from_and_change(inputs, strategy, change_config);
+        match add_change_result {
+            Ok(_) => {
+                let fee = self.get_fee_if_set().unwrap();
```
lisicky, line 1849: Try to avoid using unwrap. It is difficult to catch side it on js side

PR #698, opened 2024-10-15
13.1.0

lisicky:
Changes:

- Fixed json schema generator
- Added `.as_address()` for `PlutusData` type 
- `PlutusData.from_address()` return error for a reward address, because otherwise it might lead misinterpretation of an address.

### DIFF
--- a/rust/json-gen/src/main.rs
+++ b/rust/json-gen/src/main.rs
@@ -7,6 +7,7 @@ use cardano_serialization_lib::*;
 macro_rules! gen_json_schema {
     ($name:ident) => {
         //let out_dir = std::env::var_os("OUT_DIR").expect("no env");
+        println!("Generating schema for {}", stringify!($name));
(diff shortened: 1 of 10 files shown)


lisicky (commented):
/check

lisicky (commented):
/check

rust/json-gen/src/main.rs
```diff
@@ -7,6 +7,7 @@ use cardano_serialization_lib::*;
 macro_rules! gen_json_schema {
     ($name:ident) => {
         //let out_dir = std::env::var_os("OUT_DIR").expect("no env");
+        println!("Generating schema for {}", stringify!($name));
```
vsubhuman, line 10: Is this intentional?

lisicky: Yep

lisicky (commented):
/check

PR #700, opened 2024-11-10
[drft] drp xtndd

vsubhuman:


rust/src/protocol_types/governance/drep.rs
```diff
@@ -29,6 +30,93 @@ pub enum DRepKind {
     AlwaysNoConfidence,
 }
 
+#[wasm_bindgen]
+#[derive(Clone, Debug, Hash, Eq, Ord, PartialEq, PartialOrd)]
+pub enum DRepExtendedId {
```
lisicky, line 35: My suggestion to fully implement CIP-129 GovernanceIdentifier to not have N implementations of different things that just are subset of GovernanceIdentifier

rust/src/protocol_types/governance/drep.rs
```diff
@@ -109,23 +205,36 @@ impl DRep {
         bech32::encode(&hrp, data.to_base32()).map_err(|e| JsError::from_str(&format! {"{:?}", e}))
     }
 
+    pub fn to_bech32(&self) -> Result<String, JsError> {
+        self.internal_to_bech32("drep".to_string())
+    }
+
+    pub fn to_bech32_cip129(&self) -> Result<String, JsError> {
```
lisicky, line 212: probably we need to encode only in cip129 format and do not provide API to encode in a old format when cip129 will be merged

vsubhuman:
After additional discussing it was agreed to take a different approach

PR #704, opened 2024-11-14
13.2.0

lisicky:
# Changes

### Transaction Fee Calculation Updates
- `set_fee()` now sets a static fee that remains unchanged during change calculation
  - `TransactionBuilder.build_tx()` will emit an error if minimal possible fee exceeds specified fee
- New `set_min_fee()` method added
  - Sets minimum fee threshold
  - If minimal possible fee is lower, uses the specified minimum fee value

### Additional Validation
- Added balance validation in `build_tx()`
- Added collateral balance check in `set_total_collateral_and_return()`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "cardano-serialization-lib",
-  "version": "13.1.0",
+  "version": "13.2.0",
   "lockfileVersion": 3,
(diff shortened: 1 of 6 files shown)


lisicky (commented):
/check

vsubhuman (approved):

PR #709, opened 2024-12-30
13.2.1

lisicky:
## Fixes
Updated `FixedTransaction` type to keep tag/non-tag structures for vkey and bootstrap witnesses  in accordance with the original transaction state if vkey or bootstrap was empty before.


lisicky (commented):
/check

vsubhuman (approved):

lisicky (commented):
/check

lisicky (commented):
/check

PR #713, opened 2025-01-28
14.0.0

lisicky:
## Breaking Changes

- Removed `TxBuilderConstants` from the public API. Use `cardano-cli`, `cardano-db-sync`, or third-party APIs to retrieve the latest cost models and protocol parameters.
- The `to_bech32()` method for DRep now requires a boolean flag to specify the new CIP-129 format. Additionally, the function's output has been updated to align with the latest CIP-0005 specifications.

## Updates

- The `from_bech32()` method for DRep now supports CIP-129 and both versions of CIP-0005.
- The `TxInputsBuilder` now includes a new set of functions: `add_regular_utxo`, `add_plutus_script_utxo`, and `add_native_script_utxo`. We recommend using these functions instead of the `add_***_input` methods, as the new ones handle referenced input fees automatically.
- A new `do_not_burn_extra_change` flag has been added to the `TransactionBuilderConfigBuilder`. This flag prevents burning extra ADA when it is not possible to create a new change output. By default, this flag is unset, and you need to enable it manually if needed.
- Fixed the equality function for `MultiAsset` when the assets list is empty.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "cardano-serialization-lib",
-  "version": "13.2.1",
+  "version": "14.0.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "cardano-serialization-lib",
-      "version": "13.2.0",
+      "version": "14.0.0",
       "hasInstallScript": true,
       "license": "MIT",
       "devDependencies": {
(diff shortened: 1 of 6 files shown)


vsubhuman (approved):

lisicky (commented):
/check

PR #715, opened 2025-02-06
Add byron address type getter

lisicky:


rust/src/legacy_address/address.rs
```diff
@@ -22,54 +22,56 @@ use std::{
     fmt,
     io::{BufRead, Write},
 };
+use crate::wasm_bindgen;
 
+#[wasm_bindgen]
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Copy, Clone)]
 #[cfg_attr(feature = "generic-serialization", derive(Serialize, Deserialize))]
-pub enum AddrType {
```
slantz, line 28: does it make sense to rename abstract enum to more specific name? though I guess it will not cause any confusions.

lisicky: Otherwise we will have name collision. There are shelley addresses which contains different kind of address including byron. And the byron address also have different kinds

vsubhuman (approved):

lisicky (commented):
/check

PR #729, opened 2025-05-05
14.1.2

lisicky:
Thought for a couple of seconds
Changes:

* **Credential definition**
Change script hash and key hash definition order in credential enum to respect cardano ledger behavior 

* **CBOR parsing**
  – Swapped `assert!(len==2)` for a proper `Err(CustomError)` on invalid array length

* **Transaction builders**
  – **`with_asset_and_min_required_coin_by_utxo_cost`** refactored to:
  • take into account an output address
  • compute required ADA in one pass 

* **Withdrawals**
  – Swapped `LinkedHashMap`→`BTreeMap` for deterministic ordering
  – Updated sort logic so script-withdrawals come before key-withdrawals in accordance with ledger implementation

* **Static assertions**
  – Removed obsolete `static_assert!` in `legacy_address/cbor.rs`

* **Tests**
  • correct withdrawal sorting order

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "cardano-serialization-lib",
-  "version": "14.1.0",
+  "version": "14.1.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "cardano-serialization-lib",
-      "version": "14.1.0",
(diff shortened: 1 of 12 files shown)


lisicky (commented):
/check

lisicky (commented):
/check

PR #736, opened 2025-07-28
Update wasm-bindgen

SupernaviX:
note that this is a patch to major version 14

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -56,7 +56,7 @@ getrandom = "0.2.3"
 # wasm
 [target.'cfg(all(target_arch = "wasm32", not(target_os = "emscripten")))'.dependencies]
 serde-wasm-bindgen = "0.4.5"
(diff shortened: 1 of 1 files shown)


lisicky:
Hi @SupernaviX ! Could you explain reasons why we need to bump bindgen version ?

TristonJ:
This would also be useful for my team, where we have other cargo dependencies that depend on `wasm-bindgen ^0.2.100` (and our current only alternative is to fork `cardano-serialization-lib`, or fork those dependencies).

lisicky (commented):
/check

TristonJ:
Apologies for the delay @lisicky - I just saw this. I see that you might have sorted it out with the wasm-bindgen team [here](https://github.com/wasm-bindgen/wasm-bindgen/issues/4597). I didn't notice this since we're not actually using WASM/JS right now :sweat_smile: (just the Rust bits). Thanks!

lisicky (approved):

PR #756, opened 2026-05-25
Release/16.0.0

lisicky:
## Breaking changes

* CBOR `Int` boundary extended to the full RFC 8949 range `[-2^64, 2^64 - 1]`. `Int::from_str("-18446744073709551616")` (`-2^64`) now succeeds. As a knock-on:
  * `Int::as_negative()` returns `None` for `-2^64` (magnitude does not fit `BigNum`). In 15.0.3 it returned `Some(BigNum(0))` via a silent `as u64` cast.
  * `BigInt::as_int()` returns `Some(Int)` for `-2^64` instead of `None`.
  * `Mint::as_negative_multiasset()` no longer unwraps the magnitude cast — see the `Mint` projection entry below.

* `Mint::as_positive_multiasset()` and `Mint::as_negative_multiasset()` return `Result<MultiAsset, JsError>` instead of `MultiAsset`. `MultiAsset` amounts are `u64`-backed, so an amount whose magnitude does not fit (`-2^64`, reachable only through the deliberately permissive CBOR/JSON deserializers) can no longer be projected. In 15.0.3 the same code path could not fail: `as_negative()` truncated through `as u64` and always returned `Some`, so a `-2^64` burn silently projected to an amount of `0`. Widening `Int` and making `as_positive()`/`as_negative()` return `None` on overflow made the old `.unwrap()` reachable — a panic (an unrecoverable abort across the wasm boundary, not a catchable `JsError`) triggerable by deserialized wire data. It now returns `Err` naming the offending policy and amount. The projection is all-or-nothing — no entry is silently dropped, so a caller comparing entry counts between a `Mint` and its projection still sees parity. Callers must add `?` / `.unwrap()`.
  * Knock-on: `TransactionBuilder::get_total_input()` and `get_total_output()` (already `Result`) can now surface this error via the mint/burn projection. `add_mint_asset_and_output()` and `add_mint_asset_and_output_min_required_coin()` are unaffected in practice — `MintAssets::insert` rejects out-of-`int64` amounts before the projection runs.

* `MintAssets::insert` and `MintAssets::new_from_entry` now enforce Conway `mint = {+ policy_id => {+ asset_name => nonzero_int64}}` (also matches Mary `int64`). Any value with `|x| > i64::MAX` is rejected — this includes `-2^64` and the previously-tolerated band `(i64::MAX, u64::MAX]` ∪ `[-u64::MAX, i64::MIN)`. In 15.0.3 only zero was rejected; out-of-`int64` amounts flowed through and produced on-chain-invalid mints. Clamp into `int64` before insertion. Deserialization paths (CBOR, serde JSON) are left permissive so callers can apply their own policy.

* `CostModel::from(Vec<i128>)` replaced with `impl TryFrom<Vec<i128>> for CostModel`. Values outside the CBOR `int` range now return `Err(JsError)` instead of producing a silently-malformed `Int`. Call sites must switch to `CostModel::try_from(vec![...])?` and bring `std::convert::TryFrom` into scope on 2018-edition crates.

* `Value::checked_sub` no longer clamps multiassets. The coin path still errors on underflow. The multiasset path now returns `Err(JsError("underflow"))` when an RHS amount exceeds the matching LHS amount, or when the RHS holds a non-zero asset the LHS lacks. Use `Value::clamped_sub` (backed by `SaturatingSub`) for the old behavior.

* `Value::is_zero()` is now semantic — `true` whenever every asset amount is zero. 15.0.3 also required `multiasset.len() == 0`, so a `Value` carrying an all-zero `MultiAsset` was not considered zero.

* `Value` and `MultiAsset` arithmetic (`checked_add`, `checked_sub`, `clamped_sub`, `MultiAsset::sub`) now route through `num_traits::{CheckedSub, SaturatingSub}` and drop zero asset entries plus policies that become empty; a multiasset that ends up semantically zero is replaced with `None` in `Value`, including the case where the LHS already carried an all-zero `MultiAsset` and the RHS has none. 15.0.3 only dropped entries when `len()` literally hit zero, so any all-zero or empty-but-present `MultiAsset` carried straight through.

* `encode_json_value_to_metadatum` (BasicConversions): JSON object keys whose integer value is outside `[-2^64, 2^64 - 1]` now fall back to string encoding instead of constructing a malformed `Int`.

## Other changes

* `cbor_event` bumped `2.1.3` → `2.4.0` (additive only — new `*_sz` writer methods and the `Sz` enum; no existing API removed or changed signature).
* `Int`: inherent `checked_{add,sub,mul}`, `saturating_{add,sub}`; `From<i32 / u32 / i64 / u64 / BigNum>`, `From<&Int> for i128`, `TryFrom<i128>`, `Display`, `FromStr`. Encoder uses canonical major-1 header selection for the full range (fixes a debug-build panic on `i64::MIN` and serializes values down to `-2^64`).
* `BigInt::as_int()` recognizes the two-limb `[0, 1]` representation of `-2^64`.
* RFC 8949 / CDDL byte-exact `Int` CBOR serialization tests across every header-size boundary (1 / 2 / 3 / 5 / 9 bytes), positive and negative, including `i64::MIN` and `-2^64`. Branch-coverage tests for every `BigInt::as_int()` match arm.
* Ergonomics (numeric wrappers, `Value`, `MultiAsset`, `Assets`, `PlutusData`, Vec wrappers): additional `From`/`Default`/`Hash`/`Eq` impls, iterator traits, builder-style methods, constructor macros (`assets!`, `multi_asset!`, `plutus_list!`, `plutus_map!`, `plutus_bytes!`, `plutus_constr!`), `PlutusList::push`, `impl_btmap_wrapper`.
* Fix: post-alonzo `TransactionOutput` deserialization accepts indefinite-length maps (`bf … ff`). The trailing break byte previously failed with `UnexpectedKeyType(Special)`. Unterminated maps and maps missing the mandatory amount still error.
* Fix: `from_bech32` on hash types no longer panics on a bech32 string with a valid checksum but an invalid base32 payload length (e.g. `addr_vkh1q5x4pp8`) — it returns `JsError`. Previously `from_base32().unwrap()` aborted, which is not catchable across the wasm boundary. (#759, audit finding 2.2)
* Fix: csl compiles as a Rust dep on `wasm32-unknown-unknown` under `dont-expose-wasm` (`noop_proc_macro` reachable on every target).
* Fix: `property-test-api` and `with-bench` feature compilation; empty `MultiAsset` arithmetic case.
* Build: `wasm-bindgen` widened to `">=0.2.93, <0.3"`; the npm pipeline pins per variant via `cargo update --precise` before each build (`asm.js` → `0.2.93`, others → `0.2.103`) because binaryen's `wasm2js` rejects multi-table modules emitted from `0.2.95+`.
* CI: `cargo test --all-features --all-targets`.
* Repo: `rust/Cargo.lock` is now committed. Does not affect downstream consumers (cargo ignores a library dependency's lockfile) — it pins the dev/CI build only.

---

> [!NOTE]
> **High Risk**
> Major semver release with documented breaking changes to core serialization/value/mint APIs; the diff itself is mostly versioning and lockfiles, but consumers upgrading to 16.0.0 must adapt to those API and semantics changes.
> 
> **Overview**
> This PR **bumps the package from 15.0.3 to 16.0.0** in `package.json`, `package-lock.json`, and `rust/Cargo.toml`, and **adds `rust/Cargo.lock`** (crate `16.0.0` with resolved deps).
> 
> **Rust dependency / build wiring in the diff:** `cbor_event` **2.1.3 → 2.4.0**; `wasm-bindgen` range widened to **`>=0.2.93, <0.3`** (lockfile resolves **0.2.93**); **`dont-expose-wasm`** now pulls **`noop_proc_macro`** via `dep:noop_proc_macro`, with the same crate added as an **optional** wasm32 dependency so `wasm32-unknown-unknown` builds work under that feature.
> 
> The **npm lockfile** refresh updates dev tooling (e.g. `glob` **10.5.0**, `lodash` **4.18.1**, hoisted `minimatch` **9.0.9**) alongside the version field.
> 
> **16.0.0 breaking API and behavior changes** (mint/`Int`/`Value`/`CostModel`, semantic zero checks, etc.) are described in the PR body; they are **not** shown in this diff slice beyond the version and dependency lines above.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 032e2c6e13fa33dc16e80835ddf57796073d809d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lisicky (commented):
/check

zuzunker:
@cursor review

AmbientTea:
Hello, any updates as to when the release will be done?

zuzunker (approved):

lisicky:
Hi @AmbientTea ! I apologize for the delay, there was some very extraordinary event which delayed the release. Thanks for your contribution! 
Release is here: 
https://github.com/Emurgo/cardano-serialization-lib/releases/tag/16.0.0

PR #760, opened 2026-08-07
fix: preserve Plutus language version through JSON serde

npoxon:
`PlutusScript` serializes to a bare hex string and always deserializes via `PlutusScript::new`, which is hardcoded to Plutus V1. A V2 or V3 script therefore round-trips as V1.

```rust
let s = PlutusScript::new_v2(vec![0x4e, 0x4d, 0x01]);
let r = ScriptRef::new_plutus_script(&s);
let back: ScriptRef = serde_json::from_str(&serde_json::to_string(&r).unwrap()).unwrap();
```

```
json={"PlutusScript":"4e4d01"}
before=PlutusV2  after=PlutusV1
hash_before=f5bd2659cbe2beda7e8ffc8e1e973108770d3f41dac38b13c5ede89a
hash_after =9884df433d130ee9a6cf8653787ba015638fcc5acf044fc13725fb64
```

The blast radius is wider than the script hash. The language is the namespace byte in `PlutusScript::hash`, it is the script-type discriminator inside the `script_ref` array (1, 2 or 3, per Conway `script = [0, native // 1, v1 // 2, v2 // 3, v3]`), it picks which witness-set key (3, 6 or 7) a script lands under, and it selects the cost model that feeds `script_data_hash`. `script_data_hash` sits inside the transaction body, so a witness set that has been through JSON yields a transaction with a different body hash, not merely a reference script with a different script hash. CBOR is unaffected throughout, since it encodes the language.

## Approach

Backwards compatible rather than changing the representation outright:

- V1 still serializes to a bare hex string, exactly as before.
- A bare hex string still deserializes to V1, so previously written JSON keeps working.
- V2 and V3 serialize as `{"bytes": "...", "language": "PlutusV2"}` and deserialize from it. An explicit `"language": "PlutusV1"` is accepted too, though never emitted.

Only V2/V3 JSON output changes shape, and that output is wrong today, so nothing can be depending on it correctly.

Deserialization is a hand-written `Visitor` rather than `#[serde(untagged)]`. Untagged buffers through `Content`, which needs `Deserializer::deserialize_any`, and formats that do not support that call would have broken for every language including V1, which had no bug. Untagged also collapses every failure into "data did not match any variant"; the visitor keeps the original invalid-value message and adds `missing field` and `duplicate field` where they apply. Unknown fields are ignored, matching the derived impls used by every other JSON type in the crate.

## Two wire changes worth calling out

**Binary serde output changes shape for all three languages, with no compat path.** Formats reporting `is_human_readable() == false` now take a fixed `(bytes, language)` pair; previously they got the hex string, same as JSON. No encoding can both carry the language and keep the old shape. For a standalone script an old archive fails cleanly, but for a script embedded mid-struct the new decoder can consume the hex ASCII as the byte payload and read the following field as the language discriminant, succeeding with corrupt data. The library has never documented a binary format, so this is a change to something previously unspecified rather than to a promise, but it should not go unmentioned. `serde-wasm-bindgen` reports `is_human_readable() == true` in both directions, so `to_js_value`/`from_js_value` stay on the same branch as `to_json`.

**The generated TypeScript typings break, so this cannot ship as a 16.0.x patch.** `PlutusScriptJSON` goes from

```ts
export type PlutusScriptJSON = string;
```

to

```ts
export type PlutusScriptJSON = string | { bytes: string; language: LanguageJSON };
```

Any downstream TS that treats a script as a string stops compiling. `is_referenceable()` also flips `false` to `true`, so `PlutusScript` moves from an inlined `{"type":"string"}` to a `$ref` in every enclosing schema. The Rust API is unchanged.

Happy to split the binary branch out or drop it if you would rather keep that surface untouched.

---

> [!NOTE]
> **Medium Risk**
> Fixes incorrect script hashes on JSON round-trip for Plutus V2/V3, which can invalidate transactions built from serialized UTXOs/witnesses; V1 JSON behavior is preserved.
> 
> **Overview**
> Fixes a bug where **`PlutusScript`** JSON round-trips dropped the Plutus language: scripts were always deserialized as **V1**, so **V2/V3** scripts got the wrong **script hash** (namespace byte in `hash()`). That broke JSON paths for **reference scripts** on outputs, witness sets, and UTXOs—CBOR was unaffected.
> 
> **JSON shape (backward compatible):** **V1** still serializes as a bare hex string and bare hex still means V1. **V2** and **V3** serialize as `{"bytes": "...", "language": "PlutusV2"|"PlutusV3"}` and deserialize from that object (explicit V1 object, unknown fields ignored). Non-human-readable serde (e.g. **bincode**) uses a `(bytes, language)` pair. **JsonSchema** documents both string and object forms.
> 
> Adds regression tests for JSON/bincode round-trip, witness sets, V2 outputs, and malformed input.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 02e1fec3a2a8ec459e42f9ad61a0e2a1df9a7644. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -45,6 +45,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2dabbe35f96fb9507f7330793dc490461b2962659ac5d427181e451a623751d1"
 
 [[package]]
+name = "bincode"
+version = "1.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b1f45e9417d87227c7a56d22e471c6206462cba514c7590c09aff4cf6d1ddcad"
+dependencies = [
+ "serde",
+]
+
+[[package]]
 name = "bitflags"
 version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -70,6 +79,7 @@ name = "cardano-serialization-lib"
 version = "16.0.0"
 dependencies = [
  "bech32",
+ "bincode",
  "cbor_event",
  "cfg-if",
  "clear_on_drop",
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -70,6 +70,7 @@ getrandom = { version = "0.2.3", features = ["js"] }
 opt-level = "s"
 
 [dev-dependencies]
+bincode = "1.3.3"
 quickcheck = "0.9.2"
 quickcheck_macros = "0.9.1"
 rand_os = "0.1"
--- a/rust/src/protocol_types/plutus/plutus_script.rs
+++ b/rust/src/protocol_types/plutus/plutus_script.rs
@@ -127,8 +127,6 @@ enum PlutusScriptJson {
     },
 }
 
-const PLUTUS_SCRIPT_FIELDS: &'static [&'static str] = &["bytes", "language"];
-
 fn plutus_script_from_hex<E: serde::de::Error>(
     hex_str: &str,
     language: LanguageKind,
@@ -148,7 +146,7 @@ impl serde::Serialize for PlutusScript {
         where
             S: serde::Serializer,
     {
-        // Non-self-describing formats cannot decode the either-shape form below.
+        // Formats reporting is_human_readable() == false take a fixed pair instead.
         if !serializer.is_human_readable() {
             return (&self.bytes, Language(self.language)).serialize(serializer);
         }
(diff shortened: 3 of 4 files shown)


lisicky (approved):
Thanks @npoxon !  LGTM , I think the JSON schema will be without backward compatibility, but we will release a new major version.

lisicky (approved):
/check

npoxon (commented):
/check

npoxon:
> Thanks @npoxon ! LGTM , I think the JSON schema will be without backward compatibility, but we will release a new major version.

Thanks @lisicky! Please could you approve the queued workflow run? It's stuck atm so I can't merge.

lisicky:
@npoxon new version is released https://github.com/Emurgo/cardano-serialization-lib/releases/tag/17.0.0

PR #761, opened 2026-08-07
17.0.0: require an explicit language in PlutusScript JSON

lisicky:
Follow-up to #760, for the 17.0.0 major bump.

#760 stopped `PlutusScript` JSON from dropping the language, but kept a bare hex string as an accepted input meaning V1, so older JSON would still parse. This removes that fallback.

Note #760 was merged after 16.0.0 was cut (2026-08-07 vs a release published 2026-08-01), so it is unreleased — 17.0.0 is the first release to carry either change.

## Why remove the fallback

A bare hex string in stored JSON is not "a V1 script" — every CSL through 16.0.0 wrote *every* script that way, language discarded. Reading one back as V1 sets the wrong `script_namespace()` byte, so the `ScriptHash` and the script address come out wrong. That is exactly the failure #760 was about, and it stays silent. A parse error is the better outcome, and a major bump is where it belongs.

## What changes

JSON is one object shape for every language version:

```json
{ "bytes": "4d01000033222220051200120011", "language": "PlutusV1" }
```

Dropping the `untagged` variant removes the need for `deserialize_any`, so the hand-written `Visitor` and the `is_human_readable()` special case on both the serialize and deserialize sides collapse into a derive over one private struct — 118 lines out, 44 in.

The derive preserves what the visitor did by hand: unknown fields ignored, field order irrelevant, duplicate `bytes`/`language` rejected. Binary serde (bincode) still round-trips the language; `plutus_script_binary_serde_round_trip_preserves_language` passes unchanged.

`schemars::schema_for!(PlutusScript)` now emits a plain object with `required: ["bytes", "language"]` instead of an `anyOf` of `string` and `object`.

## Compatibility

Since no released version ever emitted the object form, 17.0.0 cannot parse JSON from 16.0.0 or earlier that contains a Plutus script — through `PlutusScripts`, `TransactionWitnessSet`, `AuxiliaryData`, `ScriptRef`, `TransactionOutput`, `TransactionUnspentOutput`, `TransactionBody`, `Transaction`, `Block`, `VersionedBlock` and their plural forms. CBOR is untouched, so re-encoding via CBOR is the migration path. Called out prominently in the release notes.

## Tests

- `plutus_script_json_v1_stays_a_bare_hex_string` → `plutus_script_json_v1_carries_its_language`
- new `plutus_script_json_rejects_a_bare_hex_string`, pinning the loud failure
- `plutus_script_json_accepts_an_explicit_v1_language` folded into the round-trip test
- `..._schema_describes_both_shapes` → `..._describes_one_shape`
- in `..._rejects_malformed_input`: a string is now `invalid type` (was `invalid value`), an array `invalid length` (was `invalid type`)

`cargo test --lib`: 754 passed, 0 failed.

## Release

`ea489e8` bumps `Cargo.toml`, `package.json`, the lockfiles and the regenerated `.js.flow` to 17.0.0. Tag `17.0.0` points at that commit — merge with a merge commit (as #756 was) so the tag stays reachable from master.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "cardano-serialization-lib",
-  "version": "16.0.0",
+  "version": "17.0.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "cardano-serialization-lib",
-      "version": "16.0.0",
+      "version": "17.0.0",
       "hasInstallScript": true,
       "license": "MIT",
       "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cardano-serialization-lib",
-  "version": "16.0.0",
+  "version": "17.0.0",
   "description": "(De)serialization functions for the Cardano blockchain along with related utility functions",
   "scripts": {
     "publish-all:prod": "node scripts/build-helper.js publish-all --env prod",
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -76,7 +76,7 @@ checksum = "5d20789868f4b01b2f2caec9f5c4e0213b41e3e5702a50157d699ae31ced2fcb"
 
 [[package]]
 name = "cardano-serialization-lib"
-version = "16.0.0"
+version = "17.0.0"
 dependencies = [
  "bech32",
  "bincode",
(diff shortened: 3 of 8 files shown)


lisicky (commented):
/check

zuzunker (approved):

zuzunker:
@lisicky , the release entry will need to be very explicit in its description about this version not being able to parse any earlier JSON that contains a serialized script in it

